PR #1, opened 2022-11-21
added new properties in Song class

yellowHatpro:
- added new properties like artist id, date modified, track number, duration,  year and disc number


akshaaatt (approved):
LGTM! Thank you 💯

PR #3, opened 2022-12-03
Explicit Theme Button and introduced Material3

07jasjeet:
1) Introduced Material3 Theming and applied theming to BrainzPlayer and Home Screen.
2) Initialized Basic Colors of the App and will be adding more in future with consultation.
3) Introduced Explicit App theme (No restart required; i.e., works on click) override in settings which works with select composable layouts for now but will be updated in future to make seamless UI experience.
4) Changes inside the colors used in composables are minimal right now but can update all UI colors once consulted with the creators of the composables.


07jasjeet:
Do I need to do anything?

akshaaatt (approved):
Awesome work! Thanks a lot for these updates. I have just added a few changes to this PR in order to fix a few things.

PR #4, opened 2022-12-05
Bug Fixes and Updated Color Scheme

07jasjeet:
Expanded material3 and fixed some bugs.

### Expansion:
1) Added mini brainzPlayer's colors to ListenBrainzTheme and mentioned at appropriate places.
2) Fixed status bar foreground [MOBILE-69]
3) Updated system navigation background and foreground.
4) Replaced onAttachFragment (deprecated function) with documentation recommend function.

### Bug 1: 
When we open the app (More visible in dark mode), the back layer content takes time to load which causes that area to be white for some time (no matter how long we keep the splash screen).
[White Background Flicker.webm](https://user-images.githubusercontent.com/98077881/205621979-f7c0ce18-210b-461c-b4c7-70e8657d595a.webm)

**After fix**

[After Flicker Fix.webm](https://user-images.githubusercontent.com/98077881/205624042-4882f97d-0516-4b70-bd96-1b316fe9ffb0.webm)

### Bug 2:
Back layer content was not taking full size of screen causing the brainzPlayer's PagerView to extend to the top.

**After Fix**


app/src/main/java/org/listenbrainz/android/presentation/features/brainzplayer/ui/BrainzPlayerActivity.kt
```diff
@@ -73,7 +73,7 @@ class BrainzPlayerActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
-            ListenBrainzTheme {
+            ListenBrainzTheme(window = window) {
```
akshaaatt, line 76: Why has this been added?

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
```diff
@@ -27,7 +30,7 @@ import org.listenbrainz.android.presentation.theme.ListenBrainzTheme
 
 @AndroidEntryPoint
 class DashboardActivity : ComponentActivity() {
-
+    @RequiresApi(Build.VERSION_CODES.TIRAMISU)
```
akshaaatt, line 33: Do we want to do this?

akshaaatt (approved):
Awesome work! Looks good to me so far 💯

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
07jasjeet: "android.Manifest.permission.READ_MEDIA_AUDIO" permission in line 45 shows lint warning if we remove this.
Although this being here doesn't interrupt with my android 10 device so I think it should be fine

app/src/main/java/org/listenbrainz/android/presentation/features/brainzplayer/ui/BrainzPlayerActivity.kt
07jasjeet: Current I couldn't find a Providable Composition Local to access the current window that the composable is being used in.
This parameter is being used to set the status bars and system navigations theming of the given window.

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
akshaaatt: Thanks for the clarification.

PR #8, opened 2023-01-02
Year in music Screen Implementations

07jasjeet:
### Implements the following Year in Music changes :
1) Top Albums Screen
2) Charts Screen (2 sub-screens)
3) Navigation animations


app/src/main/java/org/listenbrainz/android/data/sources/api/YimService.kt
```diff
@@ -8,8 +8,8 @@ import javax.inject.Singleton
 @Singleton
 interface YimService {
     
-    // https://api.listenbrainz.org/1/stats/user/(user_name)/year-in-music
-    @GET("1/stats/user/{user_name}/year-in-music")
+    // https://api.listenbrainz.org/1/stats/user/(user_name)/year-in-music/2022
+    @GET("1/stats/user/{user_name}/year-in-music/2022")
```
akshaaatt, line 12: would it be a good idea to fetch the previous year over here programmatically?

akshaaatt (approved):
Great work 💯 Looks good to me!

app/src/main/java/org/listenbrainz/android/data/sources/api/YimService.kt
07jasjeet: It would just be too much parameter passing given that requests take time already and would have to hardcode or fetch it through time libs somewhere anyway so I thought hardcoding  would be best for now.
Also Drafting this pr, just keep checking :))

akshaaatt: You can make multiple PRs. Don't need to make a lengthy PR with all the features. That would get hard to review.

07jasjeet: Yes yes, this is an intermediate PR. Will finish this project with the next one (Final PR).

akshaaatt: Totally depends on you. I only have a suggestion to keep less than 10 file changes in 1 PR for better reviews.

07jasjeet: Can't say about file changes because things keep popping up so I visit back a lot but will make sure 2-3 screens are made in this one.

akshaaatt: Awesome!

PR #10, opened 2023-01-05
Year in music Screen Implementations phase 2

07jasjeet:
### In this PR: 
1) Loopholes fixed.
2) Loading implemented.
3) Users with less listens anomaly fixed.
4) Statistics Screen Implemented.


app/src/main/res/values/colors.xml
```diff
@@ -46,4 +46,11 @@
     <!--    Background-->
     <color name="app_bg">#e0e0e0</color>
     <color name="app_bg_light">#90A4AE</color>
+
+    <!-- YIM Colors-->
+    <color name="yim_150">#FFF80729</color>
```
akshaaatt, line 51: Aren't we using Jetpack Compose for the colors?

akshaaatt (approved):
Looks good to me! Thank you so much for the awesome work here 💯

app/src/main/res/values/colors.xml
07jasjeet: My bad left it here by mistake. Was trying some stuff didn't work out.

PR #15, opened 2023-01-10
Playables are changeable now

yellowHatpro:
- playables can be changed, now we can play an album, and songs of only that album will be played unless we change the playable.

https://user-images.githubusercontent.com/75999921/211518641-54064e49-fb8e-4a0c-a628-e4e201562bcc.mp4


akshaaatt (approved):
LGTM! Thanks!

PR #24, opened 2023-01-13
Implemented Yim Share Enums and implemented tests

07jasjeet:
###Whats done :
1) Implemented enums for share function in yim.
2) Fixed fps drops by removing alpha animation and restructured lazy column of front layer in view pager to improve performance.


app/build.gradle
```diff
@@ -26,9 +26,10 @@ android {
         applicationId 'org.listenbrainz.android'
         minSdk 21
         targetSdk 33
+        versionCode 8
```
akshaaatt, line 29: Unrelated. Please remove.

app/build.gradle
```diff
@@ -162,14 +163,18 @@ dependencies {
     implementation files('./lib/spotify-app-remote-release-0.7.2.aar')
 
     //Test Setup
+    def mockito_version = '2.7.1'       // Useful for mocking context
```
akshaaatt, line 166: We do not use mockito anywhere. Why are we adding this?

app/src/main/java/org/listenbrainz/android/presentation/features/brainzplayer/ui/BrainzPlayerBackDropScreen.kt
```diff
@@ -90,9 +89,16 @@ fun BrainzPlayerBackDropScreen(
                 listenLiked = listenLiked,
                 isShuffled = isShuffled,
                 repeatMode = repeatMode,
-                modifier = Modifier.padding(paddingValues).alpha(1 - (delta / maxDelta).coerceIn(0f, 1f))
+                modifier = Modifier
+                    .padding(paddingValues)
+                    .alpha(1 - (delta / maxDelta).coerceIn(0f, 1f))     // FIXME : fps drops
+            )
+            SongViewPager(
```
akshaaatt, line 96: I don't remember what @yellowHatpro's take was on this. I remember him asking to revert these changes.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/navigation/YimShareables.kt
```diff
@@ -0,0 +1,10 @@
+package org.listenbrainz.android.presentation.features.yim.navigation
+
+enum class YimShareables(val code: String) {
```
akshaaatt, line 3: Please create 2 separate PRs for these. Thank you!

akshaaatt (changes requested):
Thanks a lot for working on this! 💯 

It is better to be a bit strict right now and make good practices a habit. It only serves us to be a better individual. Continuing on the feedback, I believe we should rework this PR and bring justice to the work you have done :)

app/build.gradle
07jasjeet: My IDE automatically adds this stuff idk what is up with it

app/build.gradle
07jasjeet: I was using it to instantiate my viewmodel because it requires application context

app/src/main/java/org/listenbrainz/android/presentation/features/brainzplayer/ui/BrainzPlayerBackDropScreen.kt
07jasjeet: Its still the same thing just a different way of writing. Will revert back.

PR #29, opened 2023-01-18
Yim sharable enums added

07jasjeet:
###In this PR:
1. Added enum class for shareable types.
2. Added some padding to playlist in BrainzPlayer viewpager.


app/src/main/java/org/listenbrainz/android/presentation/features/yim/screens/components/YimShareButton.kt
```diff
@@ -55,7 +56,7 @@ fun YimShareButton(
 private fun ShowDialog(
     onDismiss: () -> Unit,
     viewModel: YimViewModel,
-    typeOfImage: Array<String>,
+    typeOfImage: Array<YimShareables>,
```
akshaaatt, line 59: Why is this an array?

app/src/main/java/org/listenbrainz/android/presentation/features/yim/navigation/YimShareables.kt
```diff
@@ -0,0 +1,10 @@
+package org.listenbrainz.android.presentation.features.yim.navigation
+
+enum class YimShareables(val code: String) {
```
akshaaatt, line 3: We could just name it as YimShareable and omit the s.
Ik its a small thing but helps maintain the sanctity of the code.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/screens/components/YimShareButton.kt
akshaaatt: I remember the dialog being separate for the artists and tracks. But I believe we should have another approach for this rather than send an array.

akshaaatt (changes requested):
Awesome! Thank you for working as requested :) 
Just a couple more things, then we should be good to go.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/navigation/YimShareables.kt
akshaaatt: Also, what is "code" here?

07jasjeet: XD the android studio suggested me this but i went this way forcefully xD. Will do

app/src/main/java/org/listenbrainz/android/presentation/features/yim/screens/components/YimShareButton.kt
07jasjeet: Used array because its fast and dialogs are same everywhere wdym?

app/src/main/java/org/listenbrainz/android/presentation/features/yim/navigation/YimShareables.kt
07jasjeet: > Also, what is "code" here?

The string that the api accepts to get response

app/src/main/java/org/listenbrainz/android/presentation/features/yim/screens/components/YimShareButton.kt
07jasjeet: I had to pass an array because pf the segmentation of all the composables. This array also determines whether to show the share button or not. Empty array means no share button.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/navigation/YimShareables.kt
akshaaatt: Right, I was suggesting an enum of the format

```kotlin
enum class YimShareable {
    TRACKS,
    STATISTICS,
    ARTISTS,
    ALBUMS,
    DISCOVERIES,
    UNDISCOVERED
}
```
But upon investigating further, I understood why you need the code, in order to pass it to the api. Sounds okay to me but I would have rather had a switch case while processing the saveSharableImage function by passing the Shareable type as the YimShareable. Both approaches sound good to me tbh. Just wanted to share where I was coming from.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/screens/components/YimShareButton.kt
akshaaatt: Just had a look at the ShowDialog function and thought we could make it more generic. But works fine for me now.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/navigation/YimShareables.kt
07jasjeet: Yea we could have a switch case but i think it'll result in same number of lines of code. Also ill capitalize the enums!

app/src/main/java/org/listenbrainz/android/presentation/features/yim/screens/components/YimShareButton.kt
07jasjeet: The dialog has special colorScheme and also the desgin is bad. :P
Will create a component for dialogs soon which we can use Appwide.

akshaaatt: Aligned

07jasjeet: ?

akshaaatt (approved):
LGTM. Thanks!

PR #33, opened 2023-01-20
Fixed firstStartError and fixed notification of brainzplayer

07jasjeet:
Fixed firstStartError which crashed the app on first ever launch.
Fixed notification player of BrainzPlayer. Now when the app is stopped, the notification player disappears. The player reappears when app is launched and play button is hit.


akshaaatt (approved):
Thanks! LGTM 💯

PR #34, opened 2023-01-20
Mobile-96

07jasjeet:
Fixed MOBILE-96 and added app-wide dialog component.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
+++ b/app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
@@ -1,29 +1,30 @@
 package org.listenbrainz.android.presentation.features.dashboard
 
+import android.Manifest.permission
 import android.content.Intent
 import android.os.Build
 import android.os.Bundle
-import android.widget.Toast
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.compose.setContent
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.appcompat.app.AppCompatDelegate.*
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
```diff
@@ -53,21 +57,37 @@ class DashboardActivity : ComponentActivity() {
             ListenBrainzTheme()
             {
                 var isGrantedPerms by remember {
-                    mutableStateOf(false)
+                    mutableStateOf(preferencePermsGranted)
```
akshaaatt, line 60: What's the reason for duplicating this here? Can't we just use one variable here?

07jasjeet: The dialog appears for some brief moment before every launch. to solve the issue i did this. Plus this can bypass the use of checkSelfPermission by introducing a simple variable access.

akshaaatt: Alright, we do need a mutable state here.

akshaaatt (approved):
LGTM. Thanks!

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
07jasjeet: Im doing a solution waiktt

akshaaatt: Haha! Can submit another PR on top of this

PR #45, opened 2023-01-30
Converted sharedTest to library

07jasjeet:
### Problem
Unit tests and instrumented tests require some functions or utils in order to work. E.g., Mock repositories, assertion utils etc.
We used to have a directory called `sharedTest` just like test or` androidTest` which has this code. The problem was the google itself doesn't want to use this type of structure as it messes up the ide. In the newer versions, a popup will always come up saying Duplicate roots connected which just means that a single directory has two or more parent roots which are using it as a dependency. Also at the moment, `androidTest` cannot access the `sharedTest` directory (for some unknown reason) which defeats the sole reason `sharedTest` exists

### Solution
Solution to this was move all the `sharedTest` code to a separate library that we inject (not like hilt) into our app as a dependency. Now, both test and `androidTest` can access the `sharedTest` directory without giving any ide errors. This solution is google recommended.


akshaaatt (changes requested):
While we are at it, it would be great if we cleaned up the repo as well by removing CollectionRepo etc from the tests and the jsons

07jasjeet:
> While we are at it, it would be great if we cleaned up the repo as well by removing CollectionRepo etc from the tests and the jsons

Done

akshaaatt (approved):
LGTM! Thanks a lot 💯

PR #47, opened 2023-02-03
MOBILE-117

07jasjeet:
### Issue
MOBILE-117: https://tickets.metabrainz.org/browse/MOBILE-117
### Solution
Even though I was not able to reproduce the bug, this check makes sure the app is in foreground. (Between onCreate and onStop)

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/App.kt
+++ b/app/src/main/java/org/listenbrainz/android/App.kt
@@ -9,6 +9,8 @@ import android.net.ConnectivityManager
 import android.net.NetworkCapabilities
 import android.os.Build
 import android.provider.Settings
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.ProcessLifecycleOwner
 import dagger.hilt.android.HiltAndroidApp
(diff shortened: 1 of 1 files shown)


akshaaatt (approved):
Idky this issue came, but the solution looks okay to me.
Never faced the need to do this on any app I've ever made.

akshaaatt:
@07jasjeet just wanted to confirm whether this is a mutable state? If not, when the lifecycle isn't created, the Listens service wont start for that app session? Shouldn't we look to initialize it once the lifecycle is ready?

07jasjeet:
> @07jasjeet just wanted to confirm whether this is a mutable state? If not, when the lifecycle isn't created, the Listens service wont start for that app session? Shouldn't we look to initialize it once the lifecycle is ready?

Oops looks like I didn't get a mail for this one 😔
This just fail safes the app from crashing, no it is not a mutable state.
I'll try adding an observer although this issue is very rare.(non-reproducible for me)

07jasjeet:
> Idky this issue came, but the solution looks okay to me.
> 
> Never faced the need to do this on any app I've ever made.

Same feelings

PR #50, opened 2023-02-03
Yim tests and code fixes

07jasjeet:
1) Yim screens fixes.
2) Yim tests implemented with certain changes to the code that make testing easier.

NOTE: About getting sharedPreferences from repository, I think this is the way to go because not only unit tests become feasible, but also UI tests become much easier to make and maintain as we don't have to make copies of functions just to bypass sharedPreferences, we can use one Mock repo to handle it all.


app/src/main/java/org/listenbrainz/android/data/repository/YimRepository.kt
```diff
@@ -5,4 +5,8 @@ import org.listenbrainz.android.util.Resource
 
 interface YimRepository {
     suspend fun getYimData(username: String): Resource<YimData>
+    
+    fun getUsername(): String?
```
akshaaatt, line 9: We shouldn't fetch the username again, but pass it as a parameter to functions from the UI.

app/src/main/java/org/listenbrainz/android/data/repository/YimRepository.kt
```diff
@@ -5,4 +5,8 @@ import org.listenbrainz.android.util.Resource
...
+    
+    fun getLoginStatus(): Int
```
akshaaatt, line 11: Same with this. Please remove these from the repo

app/src/main/java/org/listenbrainz/android/presentation/features/components/SimilarUserCard.kt
```diff
@@ -0,0 +1,138 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import org.listenbrainz.android.R
+import org.listenbrainz.android.presentation.theme.ListenBrainzTheme
+import org.listenbrainz.android.presentation.theme.isUiModeIsDark
+import org.listenbrainz.android.presentation.theme.lb_purple
+import java.text.DecimalFormat
+
+@Preview(uiMode = UI_MODE_NIGHT_NO, showBackground = true)
```
akshaaatt, line 31: Can we show 2 previews like this?

akshaaatt: Also, can you move the previews below the view?

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
```diff
@@ -48,8 +48,7 @@ class DashboardActivity : ComponentActivity() {
             Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU -> {
                 arrayOf(
                     permission.READ_MEDIA_IMAGES,
-                    permission.READ_MEDIA_AUDIO,
-                    permission.READ_EXTERNAL_STORAGE        // TODO: Remove this if not needed.
```
akshaaatt, line 52: Where did you find the doc for this? I know we discussed this before but a doc/article link would be good here.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/YimViewModel.kt
```diff
@@ -262,7 +234,9 @@ class YimViewModel @Inject constructor(private val repository: YimRepository, @A
                 )
                 
             }catch (e: Exception){
-                Toast.makeText(context, "Something went wrong.", Toast.LENGTH_SHORT).show()
+                withContext(Dispatchers.Main){
```
akshaaatt, line 237: We shouldn't show a toast from the viewmodel. A toast is a ui update. These should strictly be added only in the activities/composables.

akshaaatt (changes requested):
Thanks a lot for the work :)

app/src/main/java/org/listenbrainz/android/data/repository/YimRepository.kt
07jasjeet: This is what I was talking about when i said move shared prefs to a separate repository. What I will do is make a complete new repositories which will contain similar prefs. This is done as to make testing flexible. Also, since shared prefs are a kind of "source of truth" or data. I feel it should go to repository. I'll be refactoring the whole code for this dw about that.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/YimViewModel.kt
07jasjeet: Agreed, took a shortcut here (works) 
:P

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
07jasjeet: READ_MEDIA_AUDIO was removed by mistake, but I tested the removal of READ_EXTERNAL_STORAGE and it works fine!

app/src/main/java/org/listenbrainz/android/data/repository/YimRepository.kt
akshaaatt: I agree with the point that shared prefs can go into another repository. We should make a viewmodel for local storage and fetch the username and login status through it and feed it through the ui to the YIMViewModel. I am happy to wait for that inclusion in this PR.

app/src/main/java/org/listenbrainz/android/presentation/features/components/SimilarUserCard.kt
07jasjeet: Yes two previews at a time! Also will do.

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
akshaaatt: It would be great if we had a doc to support this. I want to get rid of these as well but as you yourself mentioned before, don't remove them if it doesn't break.

app/src/main/java/org/listenbrainz/android/data/repository/YimRepository.kt
07jasjeet: We don't need a view-model for this. Injecting a repository that has shared Preferences on any view-model (that needs it) would solve our issue. Since shared preferences have no requirement of a lifecycle, repository would do! 
Whatcha say?

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
07jasjeet: That quote tho :P Ill look for the docs.

app/src/main/java/org/listenbrainz/android/presentation/features/yim/YimViewModel.kt
07jasjeet: Apparently now I remember that Dispatchers.Main works the same as runOnUiThread{}. Looks valid to me. If I were to wait for the function to return, it'll freeze the UI or the Main thread.

app/src/main/java/org/listenbrainz/android/data/repository/YimRepository.kt
07jasjeet: Done

app/src/main/java/org/listenbrainz/android/presentation/features/dashboard/DashboardActivity.kt
07jasjeet: Funnily enough, we don't even need permission for images because if our app is the one who writes the images to mediastore (no matter the location of the image), I'm able to access the image because it exists because of our app in the first place. m! Android

app/src/main/java/org/listenbrainz/android/presentation/features/yim/YimViewModel.kt
akshaaatt: You can usually use a LaunchedEffect to perform UI updates once the function completes in a different thread. But I'll let this slip in this PR merge.

akshaaatt (approved):
LGTM! Thanks! 💯

PR #60, opened 2023-02-16
MOBILE-114 BrainzPlayer not responding fix

07jasjeet:
### Problem
https://tickets.metabrainz.org/browse/MOBILE-114

### Solution
Apparently some heavy task was not threaded to the worker thread. Right now, the app won't freeze, but loading artists, albums etc., might take time causing their respective screens to show loading signs. Devices with no songs will show a message saying "Seems like your music library is empty."


07jasjeet:
Second Commit "Optimized Code" improves performance by this margin:
Before
Delay 1
After
Delay 2

07jasjeet:
Commit [c20c103](https://github.com/metabrainz/listenbrainz-android/pull/60/commits/c20c10371223d181736095885fac626e6f4dfc6f)

Since `AlbumsData().fetchAlbums()` function is being executed multiple, it would be better
if we save the result and hence save time thus improving performance.
This result is only called once per lifetime of the app, so creating
an object would suffice instead of caching.

app/src/main/java/org/listenbrainz/android/presentation/features/brainzplayer/ui/playlist/PlaylistViewModel.kt
```diff
@@ -22,7 +23,7 @@ class PlaylistViewModel @Inject constructor(
     val playlists = playlistRepository.getAllPlaylist()
     val playlistsCollectedFromChecklist = MutableStateFlow(mutableListOf<Playlist>())
     init {
-        viewModelScope.launch {
+        viewModelScope.launch(Dispatchers.Default) {
```
akshaaatt, line 26: How did you come to the conclusion of using the default dispatcher?

07jasjeet: Since view-model's launch don't work the way normal launch does. 
I don't know how coroutine's internal working do here but here's what I think.
All view-models access the same thread for some job and that is Dispatchers.main here.
Although all docs mention that whenever we use launch, Dispatchers.Default is called. Specifying what thread I want to do the job in worked like a charm here.

akshaaatt: These are critical things. We shouldn't take decisions on hit and trial basis.  Disk I/O operations are typically blocking operations, and we don't want to perform them on the main/UI thread. We should use Dispatchers.IO for performing this. Not just here but also for the song viewmodel.

07jasjeet: Oh no no, the internal workings of room and content resolver thread themselves to ensure light weight work. Doing stuff on default here makes sense because we are making heaving calculations which are what the actual cause of blocking here. File I/O are lighter in this case lol.

07jasjeet: For example, in my case (16GB data), file i/o accounts for merely 0.0something seconds of work, where as calculations take up 4.5 seconds. This data is based on runs before optimisations. Its pretty much instantaneous now.

akshaaatt: @07jasjeet If you have a look at the `addAlbums` function in the AlbumViewModel, it uses `AlbumsData().fetchAlbums()` which performs I/O operations while querying the MediaStore, hence we use the IO dispatcher there. Same goes for the SongViewModel. For the PlaylistViewModel, and some other places, you are right. I will make those updates.

07jasjeet: I just read about content resolver and provider in depth, docs say to use cursorLoader for queries, but since that has been deprecated and replaced with what we are using now, it seems we are on the right track.

PR #61, opened 2023-02-18
MOBILE-120 BP crashes when playing an empty playlist.

07jasjeet:
### Problem
https://tickets.metabrainz.org/browse/MOBILE-120

### Solution
Added condition when hitting play. Showed a toast to notify the user that selected playlist is empty.


akshaaatt (approved):
LGTM! Thanks 💯

PR #63, opened 2023-02-21
Major BrainzPlayer fixes

07jasjeet:
BrainzPlayer Changes:

- Users can now refresh each list separately. This allows deleted songs to get removed and new songs to be added on refresh which was not possible before.
- Introduced some performance improvements in brainzPlayer activity.
- Major Performance improvements to BrainzPlayer mini player (View Pager). Earlier whenever we opened the app or brainzPlayer, it paused the UI thread for a while, that is fixed now.
- Fixed a bug which is described here: https://tickets.metabrainz.org/browse/MOBILE-108

Miscellaneous Changes:

- Upgraded compose complier version that fixes a lot of compose bugs.
- Upgraded kotlin plugin.


app/src/main/java/org/listenbrainz/android/data/repository/ArtistRepositoryImpl.kt
```diff
@@ -34,26 +38,51 @@ class ArtistRepositoryImpl @Inject constructor(
...
+                for (artist in artists) {
+                    // We do not need to refresh cache (songsListCache) here as it already got refreshed above when we created list of albums.
+                    artist.albums.addAll(addAllAlbumsOfArtist(artist.toArtist()).map {
+                        it.toAlbumEntity()
+                    })
+                }
+            }
+        }
+        
+        songsJob.await()
```
akshaaatt, line 77: We should write callback functions rather than blocking this function in IO to proceed.

sharedTest/build.gradle
```diff
@@ -2,6 +2,7 @@ plugins {
     id 'com.android.library'
     id 'kotlin-android'
     id 'kotlin-kapt'
+    id 'org.jetbrains.kotlin.android'
```
akshaaatt, line 5: This plugin id was deprecated in favour of kotlin-android plugin id. Why add this?

akshaaatt (changes requested):
Thanks for the work! :)

app/src/main/java/org/listenbrainz/android/data/dao/PlaylistDao.kt
```diff
@@ -14,10 +14,10 @@ interface PlaylistDao {
     @Query("SELECT * FROM PLAYLISTS WHERE id LIKE :playlistId")
     fun getPlaylist(playlistId: Long): Flow<PlaylistEntity>
 
-    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    @Upsert
     suspend fun insertPlaylist(playlistEntity: PlaylistEntity)
 
-    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    @Upsert
```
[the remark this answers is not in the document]
07jasjeet: Yupp the room version that was crashing had this!

app/src/main/java/org/listenbrainz/android/data/dao/SongDao.kt
```diff
@@ -13,7 +13,13 @@ interface SongDao {
     @Query(value = "SELECT * FROM SONGS")
     fun getSongEntities() : Flow<List<SongEntity>>
 
+    @Query(value = "SELECT * FROM SONGS")
+    fun getSongEntitiesAsList() : List<SongEntity>
```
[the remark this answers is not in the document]
07jasjeet: Flow are infinite if their parent isn't destroyed, which means, everytime the user refreshes, a flow is created that never ends and subsequently, as user refreshes more, more flows are created and more threads are also created which is not very noice! Also Any code after a flow isn't executed, so thats that. And another reason is that we require flow if we need to access the value constantly or observe changes in them, here we just need the value one time (for instantaneous work).

app/src/main/java/org/listenbrainz/android/data/repository/ArtistRepositoryImpl.kt
07jasjeet: Blocking has to be done in order for the artist lists to be ready for addition. The main motive here is that we are executing both the functions above asynchronously rather than doing one job and then other which is less efficient.

Correct me if I misunderstood.

sharedTest/build.gradle
07jasjeet: Automatic kotlin upgrade by ide added this, i thought i removed everything but looks like this is still here. Will be removing this.

app/src/main/java/org/listenbrainz/android/data/repository/ArtistRepository.kt
```diff
@@ -8,7 +8,7 @@ import org.listenbrainz.android.data.sources.brainzplayer.Song
 interface ArtistRepository {
  fun getArtist(artistID: String) : Flow<Artist>
  fun getArtists(): Flow<List<Artist>>
- suspend fun addArtists(): Boolean
- suspend fun addAllSongsOfArtist(artist: Artist): List<Song>
+ suspend fun addArtists(userRequestedRefresh: Boolean = false): Boolean
```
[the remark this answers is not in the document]
07jasjeet: Good suggestion, will do!

07jasjeet:
@akshaaatt here's what I've found so far (will be editing the updates) :
1) For automatic updations, we can use [notifyChange](https://developer.android.com/reference/android/content/ContentResolver#notifyChange(java.util.Collection%3Candroid.net.Uri%3E,%20android.database.ContentObserver,%20int)) through [registerContentObserver](https://developer.android.com/reference/android/content/ContentResolver#registerContentObserver(android.net.Uri,%20boolean,%20android.database.ContentObserver)) on a foreground service.
2) Base URI to feed `registerContentObserver` is this : "content://media/external/audio"
3) Registering of observers should be done in App() I think, so we are always sure that the observer is there.
4) Unregister observer to avoid memory leaks.

PR #76, opened 2023-02-28
UI Fixes : Added music duration text & fixed player control colors 

Vishvajeet590:
**BrainzPlayer Changes**

- Added music duration text below the seek bar to convey more information to the user
- Fixed button colors to be visible in both dark and light themes.

**Changes can be seen below**
Component 3

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
@@ -271,19 +271,24 @@ fun PlayerScreen(
                     .fillMaxWidth(0.98F)
                     .padding(start = 10.dp, top = 10.dp, end = 10.dp)) {
                 val song by brainzPlayerViewModel.currentlyPlayingSong.collectAsState()
+                val songCurrentPosition by brainzPlayerViewModel.songCurrentPosition.collectAsState()
                 var duration = "00:00"
+                var currentPosition = "00:00"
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
```diff
@@ -266,6 +265,32 @@ fun PlayerScreen(
...
+            Row(verticalAlignment = Alignment.CenterVertically,
+                horizontalArrangement = Arrangement.SpaceBetween,
+                modifier = Modifier
+                    .fillMaxWidth(0.98F)
+                    .padding(start = 10.dp, top = 10.dp, end = 10.dp)) {
+                val song by brainzPlayerViewModel.currentlyPlayingSong.collectAsState()
+                var duration = "00:00"
+
+                Text(
+                    text = "00:00",
```
akshaaatt, line 277: Shouldn't this represent the current duration at which the song is when playing?

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
Vishvajeet590: Yup, missed that. Added that in the recent commit. Have look at it.

akshaaatt (approved):
LGTM! Thanks a lot for the work 💯

PR #77, opened 2023-03-02
Converted activities to Composables

07jasjeet:


akshaaatt (approved):
LGTM! Thanks for the amazing work 💯 
Do have a look at the updates made and let's follow that in the upcoming PRs.

PR #78, opened 2023-03-02
Add : Profile screen converted to Composable.

Vishvajeet590:


app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginScreen.kt
```diff
@@ -0,0 +1,206 @@
...
+private fun startLogin(context: Context) {
+    val intent = Intent(
+        Intent.ACTION_VIEW,
+        Uri.parse(
+            ListenBrainzServiceGenerator.AUTH_BASE_URL
+                    + "authorize"
+                    + "?response_type=code"
+                    + "&client_id=" + ListenBrainzServiceGenerator.CLIENT_ID
+                    + "&redirect_uri=" + ListenBrainzServiceGenerator.OAUTH_REDIRECT_URI
+                    + "&scope=profile%20collection%20tag%20rating"
```
akshaaatt, line 131: Let's remove these scopes now? We needed these in the MB android app before but don't see a need here.

akshaaatt (changes requested):

07jasjeet:
Please check out what I have done in #77 with listens activity (or BrainzPlayer activity) to get an idea what I meant and do changes as you understand. After the changes, remove the intent to start profile activity from Bottom Nav Bar, which means, the profile activity won't start when you click the profile option in bottom nav bar. I will fix it when I merge your work with mine in #77

akshaaatt:
We have to revamp the profile page entirely and populate things to it, so the effort of this PR would not be of much use if merged. Hence, closing it.

https://www.figma.com/file/tFLTQiq7QlA1oGRTJVtXml/ListenBrainz?node-id=206%3A9&t=FGkDXizks0pDLQtM-1 would be the desired design for the profile page.

PR #96, opened 2023-03-14
Feat:User Able to remove the Song from the Queue

rishuriya:
The Previous pr got some issue so I am closing it, This is the same Pr with Same functionality 
Screenshot from 2023-03-15 02-24-32

@akshaaatt Plz suggest any change if required


akshaaatt (approved):
LGTM! Thanks for the work 💯

PR #98, opened 2023-03-20
Converted Login activity to composable and fixed all tests.

07jasjeet:
### 1) Tests Fixed:

-  YimActivityTest
-  Dao Tests
-  LoginActivityTests

### 2) Converted Profile Screen to Composable.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -97,6 +97,7 @@ dependencies {
     implementation 'androidx.browser:browser:1.5.0'
     implementation 'androidx.preference:preference-ktx:1.2.0'
     implementation 'androidx.core:core-splashscreen:1.0.0'
+    //implementation 'androidx.datastore:datastore-preferences:1.0.0'
 
     //Web Service Setup
     implementation 'com.google.code.gson:gson:2.10.1'
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginActivity.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginActivity.kt
@@ -1,112 +1,115 @@
 package org.listenbrainz.android.ui.screens.login
 
-import android.content.Intent
-import android.graphics.drawable.ColorDrawable
-import android.net.Uri
 import android.os.Bundle
-import android.view.Menu
 import android.widget.Toast
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.layout.*
+import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import dagger.hilt.android.AndroidEntryPoint
-import org.listenbrainz.android.R
-import org.listenbrainz.android.databinding.ActivityLoginBinding
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.launch
 import org.listenbrainz.android.model.AccessToken
 import org.listenbrainz.android.model.UserInfo
+import org.listenbrainz.android.repository.AppPreferencesImpl
 import org.listenbrainz.android.ui.components.ListenBrainzActivity
-import org.listenbrainz.android.ui.screens.dashboard.DashboardActivity
-import org.listenbrainz.android.util.LBSharedPreferences
+import org.listenbrainz.android.ui.theme.ListenBrainzTheme
+import org.listenbrainz.android.util.LBSharedPreferences.STATUS_LOGGED_OUT
 import org.listenbrainz.android.util.ListenBrainzServiceGenerator
 import org.listenbrainz.android.viewmodel.LoginViewModel
 
 @AndroidEntryPoint
 class LoginActivity : ListenBrainzActivity() {
 
-    private var binding: ActivityLoginBinding? = null
-    private var loginViewModel: LoginViewModel? = null
+    private lateinit var viewModel: LoginViewModel
+    var isIntentLaunched : Boolean? = null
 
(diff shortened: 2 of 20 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginScreen.kt
```diff
@@ -0,0 +1,111 @@
...
+        .collectAsState(initial = viewModel.appPreferences.loginStatus, context = Dispatchers.Default)
+        .value
+    
+    Column(
+        modifier = Modifier.fillMaxSize(),
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.Center
+    ) {
+        
+        Box {
```
akshaaatt, line 55: Do we need a Box here?

app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginActivity.kt
```diff
@@ -1,112 +1,102 @@
...
-        supportActionBar!!.setDisplayHomeAsUpEnabled(true)
-        loginViewModel = ViewModelProvider(this)[LoginViewModel::class.java]
-        loginViewModel!!.accessTokenLiveData!!.observe(this) { accessToken: AccessToken? ->
-            saveOAuthToken(accessToken)
-        }
-        loginViewModel!!.userInfoLiveData!!.observe(this) { userInfo: UserInfo? ->
-            saveUserInfo(userInfo)
+    
+        // Controls launch of startLogin
+        isIntentLaunched = intent.getBooleanExtra("startLogin", false)
```
akshaaatt, line 41: Why keep this flag?

app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginScreen.kt
```diff
@@ -0,0 +1,111 @@
...
+        }
+        
+        Button(
+            onClick = {
+                when (loginStatus) {
+                    STATUS_LOGGED_IN -> {
+                        viewModel.logoutUser(context)
+                    }
+                    else -> {
+                        val intent = Intent(context, LoginActivity::class.java)
```
akshaaatt, line 71: Why are we launching an activity to handle the login flow? We could manage it right here.

akshaaatt (changes requested):
Thanks a lot for this work! 💯 

Have a few open questions before we move ahead.

app/build.gradle
```diff
@@ -97,6 +97,7 @@ dependencies {
     implementation 'androidx.browser:browser:1.5.0'
     implementation 'androidx.preference:preference-ktx:1.2.0'
     implementation 'androidx.core:core-splashscreen:1.0.0'
+    //implementation 'androidx.datastore:datastore-preferences:1.0.0'
```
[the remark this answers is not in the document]
07jasjeet: This should be our next target. It is a replacement for SharedPreferences and is part of android jetpack. It'll take time to migrate but we'll get there for sure.

app/src/main/java/org/listenbrainz/android/viewmodel/LoginViewModel.kt
```diff
@@ -1,38 +1,119 @@
...
+    fun startLogin(context: Context) {
+        val intent = Intent(
+            Intent.ACTION_VIEW,
+            Uri.parse(
+                ListenBrainzServiceGenerator.AUTH_BASE_URL
+                        + "authorize"
+                        + "?response_type=code"
+                        + "&client_id=" + ListenBrainzServiceGenerator.CLIENT_ID
+                        + "&redirect_uri=" + ListenBrainzServiceGenerator.OAUTH_REDIRECT_URI
+                        + "&scope=profile%20collection%20tag%20rating"
```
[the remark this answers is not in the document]
07jasjeet: These scopes actually cause error if absent while logging in.

app/src/main/java/org/listenbrainz/android/ui/screens/yim/YearInMusicActivity.kt
```diff
@@ -24,7 +22,7 @@ class YearInMusicActivity : ComponentActivity() {
         // Login Check
         if (!yimViewModel.isLoggedIn()){
             Toast.makeText(this, "Please Login to access your Year in Music!", Toast.LENGTH_LONG).show()
-            startActivity(Intent(this, LoginActivity::class.java))
+            //startActivity(Intent(this, LoginActivity::class.java))
```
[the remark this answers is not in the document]
07jasjeet: Actually since most of the content of profile is migrated to composable so I had to remove this.

app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginScreen.kt
07jasjeet: Since LoginAcitivity is a webview, we would need to make DashboardActivity a webview which does not count as an app as per android. Also this now being composable, other people can start working on profile screen too

akshaaatt: Jetpack compose is still behind. We need an explicit activity to be able to open the web which is not what I would have wanted. No other option anyway for now.

app/src/main/java/org/listenbrainz/android/ui/screens/login/LoginActivity.kt
07jasjeet: This didn't do it. Whenever a user clicks _No thanks_ on the site, the `code` param is again `null`, which means redirecting back to the app will again launch auth and user will be stuck in this loop, unless they close the app manually.

akshaaatt: As I had mentioned earlier to you, we want the user to authenticate the app the first thing, hence after this PR is merged, I will look to move this auth part to the onboarding page. I don't want the user to use the app if they aren't authenticated. It doesn't make sense. Given that feed will be the first page the user interacts with, the user is supposed to be authenticated.

07jasjeet: Got rid of the flag :)

akshaaatt (approved):
LGTM! Thanks 💯

PR #102, opened 2023-03-24
Feat: All Queue Features are added

rishuriya:
This PR contains the feature to add a song, Playlist, or album to the queue, and also has a feature for playing next. 
Some changes in UI such as making the card as same as the queue card because there are two different types of cards to show the list of songs and a drop-down button so the user can choose to play next or to add to the queue.

@akshaaatt bhaiya can you please review it and suggest any changes if required?

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/model/Playable.kt
+++ b/app/src/main/java/org/listenbrainz/android/model/Playable.kt
@@ -4,7 +4,8 @@ data class Playable(
     val type: PlayableType,
     val id: Long,
     val songs: List<Song>,
-    var currentSongIndex : Int
+    var currentSongIndex : Int,
+    var seekTo: Long = 0L
 )
 
 enum class PlayableType{
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
@@ -44,6 +44,7 @@ import org.listenbrainz.android.model.PlayableType
 import org.listenbrainz.android.model.Playlist.Companion.recentlyPlayed
 import org.listenbrainz.android.model.RepeatMode
 import org.listenbrainz.android.model.Song
+import org.listenbrainz.android.service.BrainzPlayerService
 import org.listenbrainz.android.ui.components.ListenCardSmall
 import org.listenbrainz.android.ui.components.PlayPauseIcon
 import org.listenbrainz.android.ui.components.SeekBar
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
```diff
@@ -371,7 +372,7 @@ fun PlayerScreen(
             }
         }
         val checkedSongs = mutableStateListOf<Song>()
-        var songs = LBSharedPreferences.currentPlayable?.songs?.toMutableList() ?: mutableListOf()
+        val songs = BrainzPlayerService.playableSongs ?: mutableListOf()
```
akshaaatt, line 375: Thanks for fixing this. We missed on this in the previous PR.

app/src/main/java/org/listenbrainz/android/model/Playable.kt
```diff
@@ -4,7 +4,8 @@ data class Playable(
     val type: PlayableType,
     val id: Long,
     val songs: List<Song>,
-    var currentSongIndex : Int
+    var currentSongIndex : Int,
+    var seekTo: Long = 0L
```
akshaaatt, line 8: What does this mean for a Playable  as the type playlist, for example?

akshaaatt:
Also, @rishuriya Add to queue doesn't seem to work right now.

rishuriya:
> Also, @rishuriya Add to queue doesn't seem to work right now.

Bhaiya, it`s working for me,  I attached a video below. can you try again if it persists then I think there will be some issues, and I will try to fix.

https://user-images.githubusercontent.com/85174423/227715638-df3ddc11-be8a-435a-9a25-cfa11e4b11f1.mp4

app/src/main/java/org/listenbrainz/android/model/Playable.kt
rishuriya: It is by default 0 even in the changePlayable function default value is given as zero, during a change in the queue list we need to use the function preparePlayer because without that users are able to see the change list in the queue but it will not play and in prepare player function is written in such a way that whenever it will be called, it start a song from the beginning, so by adding this when user change the playable list it also stores the current position of a song and use the preparePlayer function it starts a song from the where the user has listened.

akshaaatt:
Hi @rishuriya , can we connect on the metabrainz IRC channel once?

rishuriya:
> Hi @rishuriya , can we connect on the metabrainz IRC channel once?

Yess bhaiya. I am there

akshaaatt (approved):
LGTM! Although there can be improvements to the code, lets move forward for now and we will update it as we work on things more.
Thanks 💯

PR #103, opened 2023-03-25
Small fix in Favouritescreen UI (Song Title and Artist name are not displaying in light theme)

SB2318:
### Description: 

In the FavouriteSong screen, The song name and artist name are not displaying.
gsoc_prob_1

### Solution:

**Light Theme:**

gsoc_final_sol

**Dark Theme:**

gsoc_sol1

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/PlaylistScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/PlaylistScreen.kt
@@ -22,6 +22,7 @@ import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.drawscope.inset
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalContext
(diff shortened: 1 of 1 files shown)


SB2318:
@akshaaatt  Please Review it.

app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/PlaylistScreen.kt
```diff
@@ -365,11 +366,13 @@ fun OnPlaylistClickScreen(playlistID: Long) {
                     Column(Modifier.padding(start = 10.dp)) {
                         Text(
                             text = it.title,
-                            color = androidx.compose.material3.MaterialTheme.colorScheme.onSurface
+                           // color = androidx.compose.material3.MaterialTheme.colorScheme.onSurface
+                            color= Color.White
```
akshaaatt, line 370: This fix is incorrect. We should never hardcode color values for an app which supports both light mode and dark mode.

Rather, we should look to make the black colored card as white or the relevant color to adapt automatically for light mode.

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/PlaylistScreen.kt
SB2318: Okay, get it. Thank You 😇

PR #105, opened 2023-03-25
Small fix in Favouritescreen UI (Song Title and Artist name are not displaying in light theme) 

SB2318:
### Dark Theme:

UI_DARK

### Light Theme:

UI_LIGHT

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/theme/Color.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/theme/Color.kt
@@ -30,3 +30,4 @@ val yimYellow = Color(0xFFFECB49)
 val yimRed = Color(0xFFFE0E25)
 val yimWhite = Color(0xFFFEFEFE)
 val yimOffWhite = Color(0xFFF4F4F4)
+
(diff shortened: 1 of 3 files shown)


SB2318:
@akshaaatt is this right way?

app/src/main/java/org/listenbrainz/android/ui/theme/Theme.kt
```diff
@@ -109,6 +112,29 @@ fun YearInMusicTheme(
     )
 }
 
+//// For Light Color Scheme on favourite screen
```
akshaaatt, line 115: What is all this for in this PR?

app/src/main/java/org/listenbrainz/android/ui/theme/Color.kt
```diff
@@ -30,3 +30,4 @@ val yimYellow = Color(0xFFFECB49)
 val yimRed = Color(0xFFFE0E25)
 val yimWhite = Color(0xFFFEFEFE)
 val yimOffWhite = Color(0xFFF4F4F4)
+
```
akshaaatt, line 33: Avoid mutating files which don't change anything.

akshaaatt (changes requested):
From what I know, all you have to do is change the background color of the card. It doesn't look right to me that you have made so many changes.

app/src/main/java/org/listenbrainz/android/ui/theme/Color.kt
SB2318: Okay okay, I will remember this. But please give me one answer whether I reached the desire behaviour or not.

Because, These commented code is only for my uses

PR #106, opened 2023-03-26
Fav Screen Display Fixed for both light and dark mode.

SB2318:


07jasjeet (changes requested):
Please replace this whole card with ListenCardSmall. It'll automatically take care everything.

SB2318:
> While the fix looks okay to me, I do agree with @07jasjeet that we could reuse the component. @SB2318 Do you think you can take that up in this PR?

Yes, There was only one problem, previously in this app the card color and the text color was both same. for this reason, when we move onto light mode it is not updated like the card color always back. So, I only updated the package from where we take the color. You can test it, I think there you can use this.

SB2318:
> Please replace this whole card with ListenCardSmall. It'll automatically take care of everything.

Sorry, @07jasjeet I don't see your message previously. @akshaaatt please wait for some time before merging my PR. let me go through this approach also. But in both cases, we use the previously defined components.

SB2318:
### Code with ListenSmallCard:

Screenshot (32)

### Light Theme:

TRACK_LIGHT_WITH_LISTEN_SMLL

### Dark Theme:

TRACK_DARK_WITH_LISTEN_SMALL

@akshaaatt please check this.

akshaaatt:
Fine @SB2318 . Thanks for the work. Let's merge this for now and work on the refactor in another PR.

akshaaatt (approved):
LGTM! Thanks for the amazing work 💯

SB2318:
> Fine @SB2318 . Thanks for the work. Let's merge this for now and work on the refactor in another PR.

Okay

PR #117, opened 2023-03-28
Dark light mode 28032023

SB2318:
@akshaaatt Please check, whether this is the desired functionality or not. 
Here is the demo Video.

https://user-images.githubusercontent.com/87614560/228298068-2fb848db-7bd7-47e9-91a4-452ba9e875ea.mp4

and please give me a feedback, in your time.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsActivity.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsActivity.kt
@@ -69,7 +69,7 @@ class SettingsActivity : AppCompatActivity() {
             if (preference.key == PREFERENCE_SYSTEM_THEME){
 
                  val intent = Intent(this@SettingsActivity, DashboardActivity::class.java)
-                 intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK)
+                intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
                 when (newValue) {
                     "Dark" -> {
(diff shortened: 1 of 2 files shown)


SB2318:
@akshaaatt please check this issue.  But I think, some modification is needed on this PR. so, tell me whether I am able to give the required feature or not.

app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsActivity.kt
```diff
@@ -66,18 +67,28 @@ class SettingsActivity : AppCompatActivity() {
...
             // Explicit Ui Mode functionality.
             if (preference.key == PREFERENCE_SYSTEM_THEME){
+
+                 val intent = Intent(this@SettingsActivity, DashboardActivity::class.java)
+                 intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP)
                 when (newValue) {
                     "Dark" -> {
                         setDefaultNightMode(MODE_NIGHT_YES)
                         isUiModeIsDark.value = true
+                        startActivity(intent)
```
akshaaatt, line 77: Why don't we move everything below the when clause and above the return statement?

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsActivity.kt
SB2318: I want to run this two lines only when user makes some changes,like when they create a newValue(whatever it may be). see, I don't do it on else part. 
For these reason
> Why don't we move everything below the when clause and above the return statement?

PR #119, opened 2023-03-31
Theme fixed

SB2318:
https://user-images.githubusercontent.com/87614560/228992955-91722ea8-d8e8-455e-8ebe-5e087ed63eed.mp4

@akshaaatt please review

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsActivity.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsActivity.kt
@@ -16,6 +16,7 @@ import androidx.preference.SwitchPreference
 import org.listenbrainz.android.application.App
 import org.listenbrainz.android.R
 import org.listenbrainz.android.databinding.ActivityPreferencesBinding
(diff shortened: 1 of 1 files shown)


SB2318:
> Usually this means that the user loses state that they were in before switching the theme. I would recommend you to try out what reddit does on their android app. I'm pretty sure they use compose and take a little different approach. But, if not possible, this can also be accepted as an approach.

Okay, will try. Thanks for the info 😇

07jasjeet:
> @07jasjeet If you don't mind Can you please send me a recording, at least 10s.
> 
> I didn't get what is the difference in reddit. 
> 
> 
> 
> get it, will try

Don't worry btw this is prolly going to be merged, I just recommend you to look into improving this in further iterations :)

SB2318:
> > @07jasjeet If you don't mind Can you please send me a recording, at least 10s.
> > I didn't get what is the difference in reddit.
> > get it, will try
> 
> Don't worry btw this is prolly going to be merged, I just recommend you to look into improving this in further iterations :)

Thank you, actually for this I need some time. 😅
But will try

akshaaatt (approved):
LGTM! Thanks 💯

PR #130, opened 2023-04-12
LB loading animation added

07jasjeet:
Experimental animation but quite stable with very less recompositions.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/components/LBLoadingAnimation.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/components/LBLoadingAnimation.kt
@@ -1,0 +1,131 @@
+package org.listenbrainz.android.ui.components
+
+import androidx.compose.animation.core.Animatable
+import androidx.compose.animation.core.animateDpAsState
+import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.*
+import androidx.compose.runtime.*
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.drawWithCache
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/listenbrainz/android/ui/components/LBLoadingAnimation.kt
```diff
@@ -0,0 +1,127 @@
...
+ * @param rotateDuration Duration of rotate animation.
+ */
+@Preview
+@Composable
+fun LBLoadingAnimation(
+    logoSize: Dp = 60.dp,
+    middleGap: Dp = logoSize / 20,
+    sideOffset: Dp = logoSize / 4,
+    backgroundColor: Color = Color.Transparent,
+    lb_orange: Color = Color(0xFFEA743B),
```
akshaaatt, line 43: We should specify generic names for these parameters. Also, this composable could just be a LoadingAnimation.

akshaaatt (approved):
LGTM! Awesome work 💯

app/src/main/java/org/listenbrainz/android/ui/components/LBLoadingAnimation.kt
07jasjeet: These are actually supposed to be constant as is and these are the best generic names my brain could cook up 😭

akshaaatt: leftColor and rightColor could be the names @07jasjeet . The idea is to generalize components with dynamic animation options/colors, etc.

07jasjeet: oh the color params I see.

07jasjeet: Done

akshaaatt (approved):

PR #132, opened 2023-04-19
Optimised ListenService and adapted appPreferences

07jasjeet:
Listens Service was registering multiple callbacks which led to multiple submissions of the same song and multiple callbacks for same app's music notification. That is fixed now.

ClearSessions method was useless as it had no connection whatsoever with the service components itself, meaning it didn't unregister callbacks as expected.

TestAppModule established to deal with appPreferences easily and thus enabling us to use appPreferences app wide easily now.

Blocking infra added, UI pending.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -162,7 +162,7 @@ dependencies {
     implementation 'androidx.activity:activity-compose:1.7.0'
 
     // Compose Navigation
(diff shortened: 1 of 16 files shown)


akshaaatt (approved):
LGTM! Thanks for the updates! 💯

07jasjeet:
> LGTM! Thanks for the updates! 💯 

Ou
Was just going to add the UI for it.

PR #133, opened 2023-04-19
Feat: SeekBar in listenPlayer

rishuriya:
This PR includes the feature to show the progress of online playing songs in the card, also the ability to control music like seek, play, and pause from LB-android.

@akshaaatt bhaiya, can you review this and suggest any change if required.


app/src/main/java/org/listenbrainz/android/ui/screens/listens/NowPlaying.kt
```diff
@@ -9,40 +9,52 @@ import androidx.compose.material.Card
...
 import com.spotify.protocol.types.PlayerState
 import org.listenbrainz.android.R
+import org.listenbrainz.android.ui.components.SeekBar
+import org.listenbrainz.android.viewmodel.ListensViewModel
 
 @Composable
 fun NowPlaying(
     playerState: PlayerState?,
-    bitmap: Bitmap?
+    bitmap: listenPoster
```
akshaaatt, line 35: Please rename the data class as well as the parameter here.

app/src/main/java/org/listenbrainz/android/ui/screens/listens/NowPlaying.kt
```diff
@@ -124,6 +146,78 @@ fun NowPlaying(
...
     NowPlaying(
         playerState = null,
-        bitmap = null
+        bitmap = listenPoster()
     )
-}
\ No newline at end of file
+}
+@Composable
+fun ProgressBar(playerState: PlayerState?) {
```
akshaaatt, line 153: Please create a separate file for this composable.

app/src/main/java/org/listenbrainz/android/ui/screens/listens/NowPlaying.kt
```diff
@@ -124,6 +146,78 @@ fun NowPlaying(
...
+                text = duration,
+                textAlign = TextAlign.Start,
+                color = Color.White,
+                modifier = Modifier.padding(start = 5.dp)
+            )
+        }
+    }
+}
+
+data class listenPoster(
```
akshaaatt, line 220: Move this to the model folder

akshaaatt (changes requested):
Thanks for the work! 💯 
A few changes, then we should be good to go.

rishuriya:
all requested changes are done bhaiya, sorry for the delay, was a bit busy with college fest this week.

app/src/main/java/org/listenbrainz/android/viewmodel/ListensViewModel.kt
```diff
@@ -141,7 +147,10 @@ class ListensViewModel @Inject constructor(
     private fun updateTrackCoverArt(playerState: PlayerState) {
         // Get image from track
         assertAppRemoteConnected()?.imagesApi?.getImage(playerState.track.imageUri, com.spotify.protocol.types.Image.Dimension.LARGE)?.setResultCallback { bitmapHere ->
-            bitmap = bitmapHere
+            bitmap =ListenBitmap(
+                bitmap=bitmapHere,
+                id = playerState.track.uri
```
akshaaatt, line 152: I think we should have just made a separate variable here. Why store it with the bitmap like this?

rishuriya: Actually, bitmap is updated from event callback, which is called every second to get song position and to avoid updating of bitmap again and again, the function will only run when uri which it store is different from the current song.

rishuriya:
> I like the feature @rishuriya but it lags a lot for me. Can you please look into the lag?

I also faced the lag before, so I checked the CPU usage which was 20%, but I bring it down to 3 to 5%. after that I didn't face the lag, but if you encounter can you tell me if you faced it after hearing several songs or from the beginning, I am also working to create the timer inside the function so that it should not call Spotify to check the position of the song.

akshaaatt (approved):
Thanks a lot @rishuriya for this amazing functionality! 💯 
We should be moving this to the BrainzPlayer however in future.

PR #136, opened 2023-04-23
Blacklist UI added and minor tweaks

07jasjeet:
Listening Service Blacklist UI added. Users can now access the list of apps from which listens have been recorded in the past and enable/disable submitting listens from them. This option can be found on ListensScreen through the floating action button.

In minor tweaks: 
1) Front layer header height was 0.dp, changed it to 56.dp to match height of bottom navigation bar. 
2) Changed minor compose if conditions to animated visibility.


akshaaatt (approved):
LGTM! Thanks for this work 💯

PR #143, opened 2023-04-30
Complete Migration from UserPreferences to AppPreferences, Fix Listens Service and cleanup code

akshaaatt:
- The app uses AppPreferences everywhere now.
- Login flow revamped.
- Listens Service fixed. It now fetches the connected services from the web and blacklists Spotify if it is already connected on the web.
- Code cleaned up.


app/src/main/java/org/listenbrainz/android/util/ListenSessionListener.kt
```diff
@@ -26,28 +29,45 @@ class ListenSessionListener(
...
-//            override fun onResponse(call: Call<ListenBrainzExternalServices?>, response: Response<ListenBrainzExternalServices?>) {
-//                println("Services found: " + response.body())
-//                registerControllers(controllers, services = response.body())
-//            }
-//            override fun onFailure(call: Call<ListenBrainzExternalServices?>, t: Throwable) {
-//                println("Services found not found")
-//                registerControllers(controllers)
-//            }
-//        })
+        if (username != null) {
```
07jasjeet, line 38: This method gets called a lot so I would suggest making a sharedPreference and making an API call when the app starts to update it. I think we could improve this with FCM later (idkfs) but for now I think this would be the way.

akshaaatt: We don't have an API which we could automatically call to get the username. The user has to login to MB first, only then we can get the username. Once that is done, we do save it in the shared preferences and fetch it from there only. (Have a look at line 32)
FCM won't help us with this.

07jasjeet: Shouldn't it just be `if (appPreferences.username)` as its a getter property. This means username variable will always be null even if user logs on since ListensService never dies.

akshaaatt: Preferences can be nullable by default @07jasjeet

07jasjeet: No no, I mean here username variable won't change whatever the value be. This is because username variable will be runtime variable for ListenService and getters only work access them directly.

akshaaatt: Right, I get what you mean now. The fact that username variable won't be mutable in this condition, so even after logging in, the service won't start until and the unless the user kills the app and restarts.

akshaaatt: Ugh. Good catch. I don't want a viewmodel here just yet. But we might have to do it now.

akshaaatt: @07jasjeet Even with `if (appPreferences.username)` this won't work. We need an observable state here.

07jasjeet: Wait wait, its fine (I didn't unwrap the code). But still the main problem is the first argument. Come on IRC

app/src/main/java/org/listenbrainz/android/application/App.kt
```diff
@@ -24,6 +34,29 @@ class App : Application() {
                 startListenService()
             }
         }
+
+        val token = appPreferences.lbAccessToken
```
07jasjeet, line 38: I think it would be better in dashboard activity. This Application class only gets destroyed when the device switches off or user force stops the app.

akshaaatt: Yes @07jasjeet Ik that. Most probably we should have the getUser call here. We will replace this api call with the user call then in future. We can time the api call to call itself every 5 secs. That's the idea.

akshaaatt: Basically we can make the get request every 2mins or so.

akshaaatt: This is common hack when working with REST apis.

akshaaatt: Also @07jasjeet , keeping the api call in dashboard will fire it whenever the dashboard is opened which will make things inconsistent.

07jasjeet: But that does somewhat gives user the belief that they are in control of when the API is called/their internet is used. plus if the app remains dormant for a while, it'll just eat data in background.

07jasjeet: We could also show a little snackbar that says "details updated" or something to be more transparent.

07jasjeet: This topic is very two sided tbh. We could use any but the one I mentioned could be an option.
Rest is upto you to decide.

akshaaatt: The snackbar idea is good. We anyway won't be updating any visible UI changes for the user where they would need a snackbar btw.

PR #160, opened 2023-05-28
Fix: Declare NewsBrainzActivity in androidManifest

prabalsingh24:
Clicking the News button in the main screen does not open the News activity as it's not defined in the AndroidManifest

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -40,12 +40,7 @@
 
         <activity
             android:name=".ui.screens.newsbrainz.NewsBrainzActivity"
-            android:exported="false"
-            android:label="@string/title_activity_news_brainz"
-            android:theme="@style/AppTheme">
-            <meta-data
-                android:name="android.app.lib_name"
-                android:value="" />
+            android:label="@string/title_activity_news_brainz">
         </activity>
 
         <service
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -38,6 +38,16 @@
                 android:value="" />
         </activity>
 
+        <activity
+            android:name=".ui.screens.newsbrainz.NewsBrainzActivity"
+            android:exported="false"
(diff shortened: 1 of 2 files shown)


prabalsingh24:
@akshaaatt idk why but I can't seem to add you a reviewer. Tagging you :)

app/src/main/AndroidManifest.xml
```diff
@@ -38,6 +38,16 @@
                 android:value="" />
         </activity>
 
+        <activity
+            android:name=".ui.screens.newsbrainz.NewsBrainzActivity"
+            android:exported="false"
+            android:label="@string/title_activity_news_brainz"
+            android:theme="@style/AppTheme">
+            <meta-data
```
akshaaatt, line 46: This is not a shared library or anything so we don't need this tag here. Please remove it.

app/src/main/AndroidManifest.xml
```diff
@@ -38,6 +38,16 @@
                 android:value="" />
         </activity>
 
+        <activity
+            android:name=".ui.screens.newsbrainz.NewsBrainzActivity"
+            android:exported="false"
```
akshaaatt, line 43: We don't actually need to specify this until and unless an intent filter is added to the activity.

app/src/main/AndroidManifest.xml
```diff
@@ -38,6 +38,16 @@
...
+            android:label="@string/title_activity_news_brainz"
+            android:theme="@style/AppTheme">
```
akshaaatt, line 45: The default theme for the app is this, hence even if we don't duplicate it here, it will pick the same theme.

akshaaatt (changes requested):
Thanks a lot for bringing this up! 💯 
I removed it by mistake in my previous commits while removing the Settings activity.

akshaaatt:
@prabalsingh24 I was in a hurry to make a beta release of the app, hence pushed the fix for it. Thanks a lot for bringing this up though! 💯

PR #161, opened 2023-05-29
Fix: don't fetch cover art urls twice

prabalsingh24:


prabalsingh24:
@akshaaatt this makes sense right? we're already fetching urls while fetching the listens. No need to make the API call again

akshaaatt (approved):
Great point. This seems to have been missed out. Thanks a lot for looking into this and fixing 💯

07jasjeet:
This flow was used to give pre-loader URLs to work with. Since the pre loader is now removed, might as well get rid of the coverArtFlow as well and use the function as is.

akshaaatt:
@07jasjeet that sounds right to me. I think removing the flow would be the better option here @prabalsingh24 since it involves storing things separately and then accessing them via index.

akshaaatt:
For now, this is correct and I don't wanna burden the PR with optimizations because this fix is important.

PR #173, opened 2023-06-07
Use WebView to extract LB Auth token

amCap1712:
- Improve login process
- Cleanup codebase, fix NLS start and improve the FeaturesActivity


akshaaatt (approved):
Awesome work! 💯 
Thanks a lot!

PR #177, opened 2023-06-10
Setup for receiving events via WebSockets

amCap1712:
Pending:
1. ~~Reconcile local now playing with the now playing received from server.~~

BP on frontend handles this so @MonkeyDo can assist in choosing the overall strategy.


akshaaatt (approved):
Thanks a lor for this! 💯

PR #182, opened 2023-06-20
MOBILE-159: Feed section phase 1.2

07jasjeet:
# Tasks
- [x]  Setup View-model
- [x] UI for Search
- [x] Unit test for view-model

[SearchScreen.webm](https://github.com/metabrainz/listenbrainz-android/assets/98077881/fc42a995-eb5a-4742-9435-e7cdb2d5cc60)


app/src/main/java/org/listenbrainz/android/model/ResponseError.kt
```diff
@@ -0,0 +1,58 @@
...
+    
+        /** Get [ResponseError] for social type API endpoints. Automatically puts actual error message.
+         * Prefer using [getSocialResponseError] in repository functions.*/
+        private fun getSocialErrorType(apiError: ApiError) : ResponseError {
+            val error = apiError.error
+            val code = apiError.code
+            return when {
+                    code == 404 -> SocialError.USER_NOT_FOUND
+                    error?.substringAfter(' ')?.contains("is already") == true -> SocialError.ALREADY_FOLLOWING
+                    error?.substringAfter(' ') == "cannot follow yourself." -> SocialError.CANNOT_FOLLOW_SELF
```
akshaaatt, line 43: Oh god. What is happening here?
You have gone too deep here and the whole logic will start breaking if the text is changed. Remove this asap. We should have defined error codes for these scenarios.

akshaaatt (changes requested):
Thanks a lot for the amazing work! 💯 We have to further work on this PR though.

Remove all the hardcodings. I might have missed commenting at a few places.

app/build.gradle
```diff
@@ -188,6 +189,10 @@ dependencies {
     testImplementation 'com.squareup.okhttp3:mockwebserver:5.0.0-alpha.11'
     testImplementation 'androidx.arch.core:core-testing:2.2.0'
     testImplementation 'org.jetbrains.kotlinx:kotlinx-coroutines-test:1.7.1'
+    // Optional -- Mockito framework
+    testImplementation "org.mockito:mockito-core:5.3.1"
```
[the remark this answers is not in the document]
07jasjeet: I Going to use mockito to mock classes. Apparently its really good. I think creating fakes for everything won't be the way to go in future as well.

app/src/main/java/org/listenbrainz/android/model/SocialError.kt
```diff
@@ -0,0 +1,11 @@
+package org.listenbrainz.android.model
+
+enum class SocialError(override val genericToast: String, override var actualResponse: String? = null): ResponseError {
+    
+    USER_NOT_FOUND(genericToast = "User not found."),     // "User Some_User_That_Does_Not_Exist not found"
+    
+    ALREADY_FOLLOWING(genericToast = "Already following user."),     // "Jasjeet is already following user someotheruser"
+    
+    CANNOT_FOLLOW_SELF(genericToast = "Whoops, cannot follow yourself.");  // "Whoops, cannot follow yourself."
```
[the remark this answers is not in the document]
07jasjeet: Alright

app/build.gradle
```diff
@@ -135,6 +135,7 @@ dependencies {
     implementation "com.github.bumptech.glide:compose:1.0.0-alpha.1"
     implementation 'io.coil-kt:coil-compose:2.4.0'
     implementation 'com.caverock:androidsvg-aar:1.4'
+    implementation 'androidx.test.ext:junit-ktx:1.1.5'
```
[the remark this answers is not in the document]
07jasjeet: Alright

app/src/main/java/org/listenbrainz/android/viewmodel/SearchViewModel.kt
```diff
@@ -3,42 +3,141 @@ package org.listenbrainz.android.viewmodel
...
 import org.listenbrainz.android.repository.SocialRepository
 import org.listenbrainz.android.util.Resource
 import javax.inject.Inject
 
 @HiltViewModel
-class SearchViewModel @Inject constructor(private val repository: SocialRepository) : ViewModel() {
+class SearchViewModel @Inject constructor(
+    private val repository: SocialRepository,
+    private val appPreferences: AppPreferences,
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher
```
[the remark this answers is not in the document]
07jasjeet: https://developer.android.com/kotlin/coroutines/coroutines-best-practices

app/src/main/java/org/listenbrainz/android/model/ResponseError.kt
07jasjeet: Unfortunately we don't have unique error codes for these situations. Yes the code will break if something was to be changed but that's just how it is. Here I've tried to make it efficient so as to not compare the whole string. If somehow, the response is to change, UNKNOWN_ERROR will be thrown which is intended behaviour.

07jasjeet: The only way to identify the error is by the string in the response.
Maybe we can prompt lucifer to add an extra field so error detection is easy?
@amCap1712

amCap1712: Why not just display the error message directly to the user in case of `code == 400`?

07jasjeet: @amCap1712 it restricts some amount of flexibility imo.

akshaaatt: Welcome to the land of overengineering

akshaaatt (approved):
Thanks for the work! Approving for now. We can make optimizations later and move forward.

PR #194, opened 2023-07-11
MOBILE-159: Feed section phase 1.2 Wrap up

07jasjeet:
Not being tracked by an exact ticket but is an extension of MOBILE-159 and covers one small task:
- [X] Updated View-model tests.


app/src/main/java/org/listenbrainz/android/di/ServiceModule.kt
```diff
@@ -35,7 +35,6 @@ class ServiceModule {
     
     
     @get:Provides
-    @get:Singleton
```
akshaaatt, line 38: Why would you not keep the services Singleton anymore?

app/src/main/java/org/listenbrainz/android/viewmodel/SearchViewModel.kt
```diff
@@ -184,4 +190,9 @@ class SearchViewModel @Inject constructor(
             inputQueryFlow.emit("")
         }
     }
+    
+    /** True if the error states the the user is already being followed.*/
+    private fun userIsAlreadyFollowed(error: ResponseError?): Boolean =
+        error == ResponseError.BAD_REQUEST &&
+                error.actualResponse?.contains("already following") == true
```
akshaaatt, line 197: avoid this as said before

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/viewmodel/SearchViewModel.kt
07jasjeet: This is a temporary solution until we the api itself is updated or we have profile screen in the app so we can remove the follow button all together.

app/src/main/java/org/listenbrainz/android/di/ServiceModule.kt
07jasjeet: Will ping you on IRC for this one.

PR #198, opened 2023-07-18
Codebase refactor

07jasjeet:
Basic refactor (refactored snake-case to camel-case) and initialised feed project.
Check out commits for more.


app/src/main/java/org/listenbrainz/android/repository/SocialRepositoryImpl.kt
```diff
@@ -10,10 +10,8 @@ import org.listenbrainz.android.service.SocialService
 import org.listenbrainz.android.util.Resource
 import org.listenbrainz.android.util.Utils.logAndReturn
 import javax.inject.Inject
-import javax.inject.Singleton
 
 
-@Singleton
```
akshaaatt, line 16: I thought we had decided that we shall deal with the singleton removal later.

app/src/main/java/org/listenbrainz/android/ui/screens/yim/YimDiscoverScreen.kt
```diff
@@ -247,16 +253,64 @@ private fun YimTopAlbumsFromArtistsList(
...
-                coverArtUrl = getCoverArtUrl(
-                    caaReleaseMbid = newReleasesOfTopArtist[index].caaReleaseMbid,
-                    caaId = newReleasesOfTopArtist[index].caaId
-                ),
-                onClick = {},
-                errorAlbumArt = R.drawable.ic_erroralbumart
-            )
+            
+            // Listen Card
+            Surface(
```
akshaaatt, line 258: We should use a component here and not manually create the card here.

app/src/main/java/org/listenbrainz/android/ui/screens/yim/YimRecommendedPlaylistsScreen.kt
```diff
@@ -252,17 +256,66 @@ private fun YimTopDiscoveriesOrMissedList(
...
-                onClick = {
-                    context.startActivity(
-                        Intent(Intent.ACTION_VIEW, Uri.parse(item.key.identifier))
-                    )
-                },
-                errorAlbumArt = R.drawable.ic_erroralbumart
-            )
+            
+            // Listen Card
+            Surface(
```
akshaaatt, line 261: Same here

app/src/main/java/org/listenbrainz/android/ui/navigation/AppNavigation.kt
```diff
@@ -37,5 +38,8 @@ fun AppNavigation(
         composable(route = AppNavigationItem.Profile.route){
             ProfileScreen(shouldScrollToTop = shouldScrollToTop)
         }
+        composable(route = AppNavigationItem.Feed.route){
```
akshaaatt, line 41: Replace the Home screen with the Feed screen. I'll move the content from there to Explore screen

app/src/main/java/org/listenbrainz/android/ui/navigation/TopBar.kt
```diff
@@ -1,4 +1,4 @@
-package org.listenbrainz.android.ui.components
+package org.listenbrainz.android.ui.navigation
```
akshaaatt, line 1: Move this back to the components folder. This is a reusable component and not a part of the navigation.

akshaaatt (changes requested):
Thanks for the changes! 💯 

Just a few fixes here and there.

app/src/main/java/org/listenbrainz/android/ui/navigation/AppNavigation.kt
07jasjeet: Alright

app/src/main/java/org/listenbrainz/android/ui/navigation/TopBar.kt
07jasjeet: Pleaseee just let this be here because it just seems extremely intuitive to have this here. Also we don't use it anywhere else.

app/src/main/java/org/listenbrainz/android/repository/SocialRepositoryImpl.kt
akshaaatt: Cool

app/src/main/java/org/listenbrainz/android/ui/navigation/TopBar.kt
07jasjeet: xD I'm think I'm really really really really fine with this change :)))

PR #207, opened 2023-07-26
MOBILE-138: Feed Section Phase 2.1

07jasjeet:
## Tasks
- [X] Feed Repository and other functions
- [X] Feed Service
- [x] Basic UI implementation

## Surplus
Added auth interceptor for directly injecting access token without unnecessary repeated parameter passing.


gradle.properties
```diff
@@ -1,7 +1,7 @@
 org.gradle.jvmargs=-Xmx4096m
 android.useAndroidX=true
-android.enableJetifier=true
+android.enableJetifier=false
```
akshaaatt, line 3: I think we should remove this line altogether

akshaaatt (approved):
Awesome work! 💯

gradle.properties
akshaaatt: On second though, we should keep it to true as per https://stackoverflow.com/a/56490827/14790459

07jasjeet: yeah seems too risky I guess but the fast build times are awesome too. I'm deeply divided xD. Will revert it in the next PR #209.

akshaaatt: Cool

PR #216, opened 2023-08-13
Fixes: Theme changes now retain state

07jasjeet:
## What's new?
1) When switching theme, the view models won't get destroyed meaning state will be retained. This also results in quick transition and gives accurate results.

2) YIM was not launching even with user logged in. Fixed that bug (cased by me).


akshaaatt (approved):
Thanks for the updates! 💯

PR #233, opened 2023-08-24
MOBILE-142: Feed section phase 2.4

07jasjeet:
## Tasks

- [X] Dropdown for feed screens
- [X] All the dialogs for feed. (Pin, Review, PersonalRecommendation)
- [X] All API calls for dropdown.


app/src/main/java/org/listenbrainz/android/repository/feed/FeedRepository.kt
```diff
@@ -1,23 +1,23 @@
...
+import org.listenbrainz.android.model.feed.FeedData
+import org.listenbrainz.android.model.feed.FeedEventDeletionData
+import org.listenbrainz.android.model.feed.FeedEventVisibilityData
 import org.listenbrainz.android.model.SocialResponse
 import org.listenbrainz.android.util.Resource
 
 interface FeedRepository {
     
-    suspend fun getFeedEvents(username: String, maxTs: Int? = null, minTs: Int? = null, count: Int = 25) : Resource<FeedData>
+    suspend fun getFeedEvents(username: String?, maxTs: Int? = null, minTs: Int? = null, count: Int = 25) : Resource<FeedData>
```
akshaaatt, line 11: Why make username nullable at all?

app/src/main/java/org/listenbrainz/android/repository/feed/FeedRepositoryImpl.kt
```diff
@@ -15,12 +15,14 @@ class FeedRepositoryImpl @Inject constructor(
...
     
     override suspend fun getFeedEvents(
-        username: String,
+        username: String?,
         maxTs: Int?,
         minTs: Int?,
         count: Int
     ) : Resource<FeedData> =
         runCatching {
+            if (username.isNullOrEmpty()) return@runCatching Resource.failure(error = ResponseError.AUTH_HEADER_NOT_FOUND)
```
akshaaatt, line 24: Perform this check before and don't include this in the repos

app/src/main/res/values/strings.xml
```diff
@@ -229,4 +229,5 @@
     <string name="tt_yim_charts_parent">tt_yim_charts_parent</string>
     <string name="tt_yim_statistics_parent">tt_yim_statistics_parent</string>
     <string name="title_activity_settings">SettingsActivity</string>
+    <string name="cb_review_tc">You acknowledge and agree that your contributed reviews to CritiqueBrainz are licensed under a Creative Commons Attribution-ShareAlike 3.0 Unported (CC BY-SA 3.0) license. You agree to license your work under this license. You represent and warrant that you own or control all rights in and to the work, that nothing in the work infringes the rights of any third-party, and that you have the permission to use and to license the work under the selected Creative Commons license. Finally, you give the MetaBrainz Foundation permission to license this content for commercial use outside of Creative Commons licenses in order to support the operations of the organization.</string>
```
akshaaatt, line 232: Why are we still using the strings from res? Just add these texts to the Constants file

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/repository/feed/FeedRepository.kt
07jasjeet: If the user isn't logged in, this field would be null

app/src/main/java/org/listenbrainz/android/repository/feed/FeedRepositoryImpl.kt
07jasjeet: Its just better to have these checks here because then i dont have to repeat this in 10 different view models. A bit anti-societal but i think im done with google's guildlines.

app/src/main/res/values/strings.xml
07jasjeet: IDE recommendation. Plus this is a long text and standalone objects acquire runtime memory and persists application lifecycle.

07jasjeet: Given that this isn't going to be frequently used as much. I think it should belong here.

app/src/main/java/org/listenbrainz/android/repository/feed/FeedRepository.kt
akshaaatt: I mean.. that was understandable already, I just wanted to say that why even check for the feed of a signed-out user?

app/src/main/java/org/listenbrainz/android/repository/feed/FeedRepositoryImpl.kt
akshaaatt: Okay tbh I don't oppose this. Let's let it be like this

app/src/main/res/values/strings.xml
akshaaatt: I wanna get away from these stupid config files that android gives us. Let the devs create their own files and follow the process their way, I feel. But for now, let's let it be then. Also, the memory reason isn't good enough. We want better code readability. Memory consumption for such text is really irrelevant

app/src/main/java/org/listenbrainz/android/repository/feed/FeedRepository.kt
07jasjeet: That is something I think we were going to figure out in future on how to ONLY allow a user when they are logged in.
On the other hand, this repository function should be foolproof in itself and not rely on forward implementations as its an isolated layer in itself.

app/src/main/res/values/strings.xml
07jasjeet: I think I might know of a way

akshaaatt (approved):
Thanks a lot for the amazing work! 💯

PR #235, opened 2023-08-27
Bug fixes

07jasjeet:
## What's fixed?
1) Pagers on all screens now handle edge conditions better.
2) Settings screen's navigation was buggy.
3) PinFeedLayout showed blurbContent section even if that section was empty.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/feed/FollowListensPagingSource.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/feed/FollowListensPagingSource.kt
@@ -4,9 +4,9 @@ import androidx.paging.PagingSource
 import androidx.paging.PagingState
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
+import org.listenbrainz.android.model.ResponseError
 import org.listenbrainz.android.model.feed.FeedData
 import org.listenbrainz.android.model.feed.FeedEventType
-import org.listenbrainz.android.model.ResponseError
 import org.listenbrainz.android.repository.feed.FeedRepository
 import org.listenbrainz.android.util.Resource
 
@@ -17,11 +17,8 @@ class FollowListensPagingSource(
     private val ioDispatcher: CoroutineDispatcher
 ): PagingSource<Int, FeedUiEventItem>() {
     
-    override fun getRefreshKey(state: PagingState<Int, FeedUiEventItem>): Int? {
-        return if ((state.anchorPosition ?: 0) < 10)
-            null
-        else
-            state.lastItemOrNull()?.event?.created
+    override fun getRefreshKey(state: PagingState<Int, FeedUiEventItem>): Int {
+        return (System.currentTimeMillis()/1000).toInt()
     }
     
     override suspend fun load(params: LoadParams<Int>): LoadResult<Int, FeedUiEventItem> {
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/feed/FollowListensPagingSource.kt
```diff
@@ -17,11 +17,8 @@ class FollowListensPagingSource(
     private val ioDispatcher: CoroutineDispatcher
 ): PagingSource<Int, FeedUiEventItem>() {
     
-    override fun getRefreshKey(state: PagingState<Int, FeedUiEventItem>): Int? {
-        return if ((state.anchorPosition ?: 0) < 10)
-            null
-        else
-            state.lastItemOrNull()?.event?.created
+    override fun getRefreshKey(state: PagingState<Int, FeedUiEventItem>): Int {
+        return (System.currentTimeMillis()/1000).toInt()
```
akshaaatt, line 21: This doesn't seem like the best solution.

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/ui/screens/feed/FollowListensPagingSource.kt
akshaaatt: `The getRefreshKey method of a PagingSource is used to determine which page key should be used if a refresh is triggered (i.e., when the data gets invalidated and needs to be loaded from the start).` How is the solution relevant?

07jasjeet: it is. This is just the key that will be given to pager when we do swipe to refresh. Also, Server takes time in milliseconds

07jasjeet: Every time we do swipe to refresh, this function will be called and all the paging data reset.

akshaaatt: Oh so the server takes the nearest timestamp and returns the feed accordingly? I was of the opinion it would take in an exact timestamp

akshaaatt: Okay the `max_ts` here would be this then

07jasjeet: It just does ts <= max_ts on the backend

akshaaatt: Cool

akshaaatt (approved):
Thanks! 💯

app/src/main/java/org/listenbrainz/android/ui/screens/feed/FollowListensPagingSource.kt
07jasjeet: Actually, providing null would also do the job since null means the server will take current time automatically.

07jasjeet: Ouu 10 seconds too late.

PR #245, opened 2023-09-02
Feed section phase 2.5

07jasjeet:
## Tasks

- [x] General improvements
- [ ] Unit tests (Next PR)
- [ ] UI Tests (Next PR)


sharedTest/src/main/java/org/listenbrainz/sharedtest/utils/ResourceString.kt
```diff
@@ -38,6 +41,30 @@ object ResourceString {
         EntityTestUtils.loadResourceAsString("user_does_not_exist_error.json")
     }
     
+    val my_feed_page_1 by lazy {
+        EntityTestUtils.loadResourceAsString("my_feed_page_1.json")
```
akshaaatt, line 45: Can we make the mock files more descriptive and not keep them in the _number format?

app/src/main/java/org/listenbrainz/android/service/ApiService.kt
```diff
@@ -0,0 +1,148 @@
...
+import org.listenbrainz.android.util.Constants.Headers.AUTHORIZATION
+import retrofit2.Response
+import retrofit2.http.Body
+import retrofit2.http.GET
+import retrofit2.http.Header
+import retrofit2.http.POST
+import retrofit2.http.Path
+import retrofit2.http.Query
+
+interface ApiService {
```
akshaaatt, line 30: Why fit everything into one single service? I don't think this is a good thing to do. Please state your reasons to do so, otherwise revert

app/src/main/java/org/listenbrainz/android/ui/components/dialogs/BaseDialog.kt
```diff
@@ -8,6 +8,8 @@ import androidx.compose.foundation.layout.ColumnScope
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.sizeIn
+import androidx.compose.foundation.rememberScrollState
```
akshaaatt, line 11: @07jasjeet, please remember that all these components should be moved to the design system soon. So keep them as generic as possible.

akshaaatt:
Also, the YearInMusicActivityTest seems to be breaking for me. Wonder why the CI didn't fail for this though.

sharedTest/src/main/java/org/listenbrainz/sharedtest/utils/ResourceString.kt
07jasjeet: This is as descriptive as it gets since these are results of one endpoint but with different parameters (here, max_ts).

app/src/main/java/org/listenbrainz/android/ui/components/dialogs/BaseDialog.kt
07jasjeet: Understood.

07jasjeet:
> Also, the YearInMusicActivityTest seems to be breaking for me. Wonder why the CI didn't fail for this though.

I think I may have changed some things here and there but I don't exactly know what errors you might be getting.

app/src/main/java/org/listenbrainz/android/service/ApiService.kt
akshaaatt: How are you measuring the performance differences? We can't just go by words because theoretically speaking, "It's accurate that creating a Retrofit instance is relatively expensive in terms of performance, as it involves setting up a series of internal components, parsing annotations, and often establishing network configurations. However, the term "heavy" is relative. In most scenarios, it's a one-time cost that doesn't noticeably affect the end-user experience, especially if done during the app's initialization phase." 

That being said, I do not want the profile/dashboard services in a single file with feed services. It just doesn't make sense to pollute the codebase by merging everything into a single file. Keep those separate. If the app performance has not taken a huge toll and we have numbers to prove it, we should make sure we don't end up confusing the dev codebase for it.

akshaaatt: The different base urls reason is invalid btw.

07jasjeet: Well for once, LB website does it the same way, and having more Singletons have actual performance impact. For the Creation part, the heavy retrofit instances do impact the first time launch (crucial) but also take up chunks of memory without serving no actual use. These different retrofit services would serve us only if we had the microservices type architecture where codebase and endpoint is different for each service.

I'm just stating facts provided by the actual library docs and most of the community agrees.

07jasjeet: Also it isn't that confusing tbh. Its more precise such that a dev will automatically know where to look an API call when they need it.

akshaaatt: We have added less than half of the endpoints we need to add for stuff like taste, recommendations, playlists, etc. We can't have everything in one file. I would go for better code abstraction over meaningless performance improvement here.

akshaaatt: You think following a comment-based approach is better in code, but I disagree. Your code should speak for itself and trust me devs hardly read in-code documentation. Any file with too much code is just a pain to deal with. That is why we always like to prefer break things into components wherever possible. Also, don't take the LB website to the perfect example of a codebase. It is still nascent.

07jasjeet: Its okay tbh. Sometimes when the apps get big, abstractions only make things difficult. And the performance improvement is not negligible. If we keep this up, its going to bite us in future for sure.

Also, the interface is barely 160 lines long. People are usually fine with 1000+ loc as long as ctrl+f is there.

akshaaatt: @07jasjeet are you familiar with SOLID principles? Please read through those once. I would like us to follow only those and avoid conflicts. Also, please set up a performance monitor like Android profiler to suit arguments with numbers from next time. I am not trying to make the app less performant but being a developer, having a single ApiService makes me want to drive away from the codebase as far as possible, hence the resistance.

07jasjeet: Also, sacrificing app performance and memory usage for Developer experience is kinda absurd to me. If people can deal with it on ListenBrainz Server idk why we should soften things up.

07jasjeet: We don't need to force principles it'll make things more confusing for newcomers plus I'm not doing anything out of the world here. Its just that this is an Android-specific case where such principles don't apply. Those principles are generalised for every technology but there are instances where you have to do what you have to do.

07jasjeet: Well, for performance, its just the first time, not that much. But the main point is unnecessary memory occupation which will last application lifecycle.

07jasjeet: Since retrofit uses okHttp, here a quote from them:
> OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.

akshaaatt: We can create a singleton of OkHttpClient to create multiple Retrofit instances. Just pass in the already initialized client within the `Retrofit.Builder().baseUrl("abc").client(okHttpClient)`

akshaaatt: > Here's the LB server [file](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/utils/APIService.ts) I'm talking about.

This is really bad code. I might talk with the team about it soon enough. Don't reference the LB code too much because it is really not the best. A lot of things are hacks.

akshaaatt, line 30: @07jasjeet can you remove this file? Also, just ping me when to review this PR.

07jasjeet: Mb forgot this one

07jasjeet: Ready to review @akshaaatt

app/src/main/java/org/listenbrainz/android/viewmodel/BaseViewModel.kt
```diff
@@ -0,0 +1,30 @@
+package org.listenbrainz.android.viewmodel
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.launch
+import org.listenbrainz.android.model.ResponseError
+
+abstract class BaseViewModel<UiState>: ViewModel() {
```
akshaaatt, line 10: abstract classes are a pain if we don't handle them properly. They end up being a waste of time in future. Just be sure to use it everywhere because it could end up biting you in the future. That being said, I think it fits well here rn and I don't see any downside of it right now so okay with it.

07jasjeet: I was thinking to delegate all these tasks of vm to a special vm only for this task app wide. What do you say?

akshaaatt: This sounds better and good to me. Separating vms is better than making all vms conform to the same methods

07jasjeet: Done

akshaaatt:
@07jasjeet can you fix the tests for this? Also, ready for review right?

07jasjeet:
> @07jasjeet can you fix the tests for this? Also, ready for review right?

Oops didn't knew tests broke again gimme a sec.

PR #317, opened 2024-01-02
Added Beta Build Type

07jasjeet:


app/build.gradle
```diff
@@ -47,33 +47,25 @@ android {
...
             versionNameSuffix ".debug"
         }
+        beta {
+            debuggable true
+            addLocalProperties(beta, localPropertiesFile)
+
+            buildConfigField "String", "API_BASE_URL", '"https://beta-api.listenbrainz.org/1/"'
+            buildConfigField "String", "WEBSITE_BASE_URL", '"https://beta.listenbrainz.org/"'
+
+            applicationIdSuffix ".beta"
```
akshaaatt, line 65: This defeats the whole purpose of us making this separation. We need to create separate environments and not builds. We want to be able to select the beta configuration and deploy that app on the playstore for a beta. The app shouldn't say ListenBrainz.beta when downloaded from the playstore.

app/build.gradle
```diff
@@ -47,33 +47,25 @@ android {
...
+            buildConfigField "String", "WEBSITE_BASE_URL", '"https://listenbrainz.org/"'
 
             applicationIdSuffix ".debug"
             versionNameSuffix ".debug"
         }
+        beta {
+            debuggable true
+            addLocalProperties(beta, localPropertiesFile)
+
+            buildConfigField "String", "API_BASE_URL", '"https://beta-api.listenbrainz.org/1/"'
```
akshaaatt, line 62: Why are we hardcoding fields in the gradle file? We should just pass a common field and then use that in the code to separate the values. The gradle file is already bloated and we shouldn't do this here at all.

app/build.gradle
```diff
@@ -117,6 +113,32 @@ android {
     }
 }
 
+static def addLocalProperties(buildType, localPropertiesFile) {
```
akshaaatt, line 116: How is thi function helping out? The local properties should be consistent and not change for different envs. Local properties are only for local dev work and shouldn't be used this way. We already replace these values while making a prod/beta build from the gh workflows I wrote. This use case is not useful for us.

akshaaatt (changes requested):
@07jasjeet please go through the comments and let me know if you have any questions. We should either redo this work or close this PR.

app/build.gradle
07jasjeet: If we are going to define a type under buildType {}, It will turn out as either one of these. We are not going to use this build anywhere other than local development. 

Usually it's recommended to go with productFlavour for this but that idea was trashed.

If you want to make releases using this build we can remove the beta suffix.

app/build.gradle
07jasjeet: We could move it to gradle.properties if it were to be changed but because this will not change, we can keep it here for now.

app/build.gradle
07jasjeet: It's the same thing as debug build, since beta build will be used for local development only, this function is just a code reduction measure, nothing else. The functionality remains similar.

akshaaatt: What makes you think that we want to use beta builds only for local testing? We should release stuff on the go and when on beta builds, use beta apis and not hit prod server.

07jasjeet: Some feature are limited in beta. Such as socket connections.

akshaaatt: If there are limitations in beta then we need to ask the backend team to resolve those. We can't have a build with some apis hiting the beta apis and some hitting the prod apis. We need to maintain consistency

akshaaatt:
Closing this PR for now. Will do it separately. Thanks a lot for getting this started though! 💯

PR #327, opened 2024-01-06
MOBILE-182:App Crashes in YIM 23

pranavkonidena:
# Problem

While scrolling to the end in any of the lists present in YIM23, the app suddenly crashes

# Solution
While creating the models , I mistakenly put caaId as string. So now what happens is when caaId is null for a song, the .toLong() function raises JavaNumberFormatException. To overcome this, I changed the datatype in the model and made some changes where it was used according to the new datatype (Long)
# Action
Please test this locally once just to make sure

# Demo of Change

https://github.com/metabrainz/listenbrainz-android/assets/122373207/93c56eb7-a8ab-41d0-88c7-d8a1f93fdb6c


07jasjeet:
No crashes, lgtm.

PR #328, opened 2024-01-07
Yim23/fixes

pranavkonidena:
This PR aims to fix all the reviews of YIM PR #311 
New Look of certain components such as heatmap and graph ->
Screenshot 2024-01-07 at 11 27 26 AM
Screenshot 2024-01-07 at 11 27 34 AM

PR #329, opened 2024-01-07
Yim23/design fixes

pranavkonidena:
# Don't Merge till #328 is merged

This PR incorporates design fixes as pointed out by aerozol here (https://docs.google.com/document/d/1XLu5srnP6j5ehW6CKtLPrhxmLWNpxG_eHeKV1W9K22k/edit)


app/src/main/java/org/listenbrainz/android/ui/screens/yim23/Yim23LandingScreen.kt
```diff
@@ -103,10 +104,10 @@ fun Yim23HomeScreen(
                                     Image(
                                         painter =
                                         painterResource(id = when (viewModel.themeType.value) {
-                                            0 -> R.drawable.yim23_pick_color_green
```
07jasjeet, line 106: In the `Yim23ThemeData` (Renamed to Yim23ThemeType) enum class, create a parameter with name `pickColorRes` and use `viewModel.themeType.pickColorRes` everywhere instead of running a when clause every time.
add other fields as well like `colorRes`, `logoRes` etc.

app/src/main/java/org/listenbrainz/android/ui/screens/yim23/Yim23LandingScreen.kt
```diff
@@ -103,10 +104,10 @@ fun Yim23HomeScreen(
...
-                                            1 -> R.drawable.yim23_pick_color_red
-                                            2 -> R.drawable.yim23_pick_color_blue
-                                            3 -> R.drawable.yim23_pick_color_grey
+                                            Yim23ThemeData.greenTheme -> R.drawable.yim23_pick_color_green
```
07jasjeet, line 105: Rename the enum class and member themes as such:
```
enum class Yim23ThemeType(@DrawableRes val colorRes: Int) {
  GREEN(pickColorRes = R.drawable.yim23_pick_color_green, colorRes = ...),
  RED(...),
  BLUE(...),
  GREY(...)
}
```

07jasjeet (changes requested):

app/src/main/java/org/listenbrainz/android/viewmodel/Yim23ViewModel.kt
```diff
@@ -30,7 +31,7 @@ import javax.inject.Inject
 class Yim23ViewModel @Inject constructor(
     private val repository: Yim23Repository,
     private val appPreferences: AppPreferences
-) : BaseYimViewModel() {
+) : ViewModel() {
```
07jasjeet, line 34: Use `BaseViewModel` here instead.

app/src/main/java/org/listenbrainz/android/ui/screens/yim23/Yim23AlbumsListScreen.kt
```diff
@@ -35,30 +23,24 @@ fun Yim23AlbumsListScreen (
...
-                    horizontalAlignment = Alignment.Start , modifier = Modifier.fillMaxWidth()) {
-                    val topReleases : List<TopReleaseYim23>? = remember {
-                        viewModel.getTopReleases()?.toList()
-                    }
-                    for(i in 1..5)
-                        Text(topReleases!![i-1].releaseName ,
-                            color = MaterialTheme.colorScheme.background ,
-                            style = MaterialTheme.typography.labelLarge , maxLines = 1)
+    Yim23BaseScreen(
+        viewModel     = viewModel,
```
07jasjeet, line 31: I like this, but for parameters, 1 space separation should be followed.

app/src/main/java/org/listenbrainz/android/viewmodel/SocialViewModel.kt
```diff
@@ -38,6 +42,15 @@ class SocialViewModel @Inject constructor(
     private val remotePlaybackHandler: RemotePlaybackHandler,
     @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
 ): FollowUnfollowModel<SocialUiState>(repository, ioDispatcher) {
+
+    var friendsData : MutableState<
```
07jasjeet, line 46: Nope delete this.

app/src/main/java/org/listenbrainz/android/viewmodel/SocialViewModel.kt
```diff
@@ -171,11 +184,17 @@ class SocialViewModel @Inject constructor(
         }
     }
 
-   fun getFollowers(username : String) : Resource<SocialData> = runBlocking {
-        val deferredResult = async{
-            repository.getFollowers(username)
-        }
-       return@runBlocking deferredResult.await()
+   fun getFollowers() {
```
07jasjeet, line 187: We need this function to be synchronous and have UI do state management on its own since this function won't be used everytime this View model is created.
```suggestion
   suspend fun getFollowers(): Resource<SocialData> {
           val username = appPreferences.getUsername()
           return repository.getFollowers(username).also {
               if(it.status == Resource.Status.FAILED){
                   emitError(result.error)
               } 
           } 
    }
```
Some of the usages of `appPreferences.getUsername()` are old and use `withContext(ioDispatcher)` but the approach is actually redundant.

07jasjeet (changes requested):

app/src/main/java/org/listenbrainz/android/ui/screens/yim23/Yim23LandingScreen.kt
pranavkonidena: Ya thats so better.. will do that thx

07jasjeet:
@pranavkonidena Few Issues: 
1) I still see crashes when scrolling down a list.
2) First screen is messed up for Devices with longer length.
3) Music Buddies screen 1 has different enter/exit animation than rest.
4) Both Music Buddies screens show weird loading signs that should not be there. Consider pre-loading followers and similar users in your `Yim23ViewModel` because your implementation demands so but you must not touch `SocialViewModel` because its a general use view-model.
5) Top and Bottom bars are still a mess.

app/build.gradle
```diff
@@ -250,4 +250,7 @@ dependencies {
     // Third party libraries
     implementation 'com.github.a914-gowtham:compose-ratingbar:1.3.4'
     implementation 'com.github.akshaaatt:Logger-Android:1.0.0'
+
+    //Charting Library (Vico)
+    implementation("com.patrykandpatrick.vico:compose:1.13.1")
```
akshaaatt, line 255: Can we please give an attribution to this library on the settings page? I think we already attribute the animations and icons we use there

akshaaatt (changes requested):

akshaaatt (commented):
@pranavkonidena please never merge main into your PRs. Although it shouldn't do anything but you shouldn't be concerned about doing so

pranavkonidena:
Updated video for YIM23 ->
https://usercontent.irccloud-cdn.com/file/qtYqQMnu/IMG_7078.MP4

07jasjeet:
More work has to be done but looks fine for beta.

07jasjeet:
@akshaaatt your call if we should fix things here on in next PR.

akshaaatt (approved):
LGTM for now. Let's fix things in the next PR

PR #336, opened 2024-01-14
MOBILE-160,184: Listen Scrobble fixes and improvements

### LINKED ISSUE #331: don't automatically scrobble any player
I've been having to clean up my listens because the android app automatically scrobbles from any new player and it had been spamming my listens with KDE connect's player notifications of my computer playing where it is already scrobbling and in a better way (not using the app's name as the artist)

But most eerily, I just noticed today when I went again to check what apps was it scrobbling from, **it was scrobbling "Phone calls" from the Phone app and Signal** (shown as its package name, org.thoughtcrime.securesms)

07jasjeet:
## Summary
Now onwards, whenever a new media app is discovered, it is opted out by default. Only user can opt-in the newly discovered media app. This helps avoid unintended scrobbles with newly discovered apps.
Migration from Blacklist to Whitelist has been handled.


akshaaatt (commented):
Thanks for working on this PR but we should have discussed this before implementing. We simply cannot move to whitelisting. This changes the UX entirely and we were not in the favour of doing this from the beginning. We want to reduce the number of steps the user takes to start submitting listens. We need to have a discussion on this properly before moving ahead with this work.

akshaaatt:
@07jasjeet why is this PR made against `main` and not `dev`?

07jasjeet:
> @07jasjeet why is this PR made against `main` and not `dev`?

Oops, forgot to change it.

app/src/main/java/org/listenbrainz/android/application/App.kt
```diff
@@ -55,8 +53,8 @@ class App : Application(), Configuration.Provider {
         MainScope().launch {
             if(
                 appPreferences.isNotificationServiceAllowed &&
-                appPreferences.getLbAccessToken().isNotEmpty() &&
-                withContext(Dispatchers.IO) { appPreferences.submitListens }
+                appPreferences.lbAccessToken.get().isNotEmpty() &&
+                appPreferences.isScrobblingAllowed.get()
```
akshaaatt, line 57: We decided on IRC to not use the term Scrobble. I think this is a good time to fix this.

app/src/main/java/org/listenbrainz/android/repository/preferences/AppPreferencesImpl.kt
```diff
@@ -44,9 +50,37 @@ import org.listenbrainz.android.util.Constants.Strings.USERNAME
 import org.listenbrainz.android.util.LinkedService
 import org.listenbrainz.android.util.TypeConverter
 
-class AppPreferencesImpl(private val context : Context): AppPreferences {
-    
+class AppPreferencesImpl(private val context: Context): AppPreferences {
     companion object {
+        private val gson = Gson()
+        private val blacklistMigration: DataMigration<Preferences> =
```
akshaaatt, line 56: Discussed on IRC to use other terms

app/src/main/java/org/listenbrainz/android/service/ListenScrobbleService.kt
```diff
@@ -29,17 +30,23 @@ class ListenScrobbleService : NotificationListenerService() {
...
             this,
             NotificationManager::class.java
         )!!
     }
-
+    private val sessionManager: MediaSessionManager by lazy {
+        ContextCompat.getSystemService(
+            this,
+            MediaSessionManager::class.java
+        )!!
```
akshaaatt, line 47: This could cause a crash

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/application/App.kt
07jasjeet: Alright

app/src/main/java/org/listenbrainz/android/repository/preferences/AppPreferencesImpl.kt
07jasjeet: I don't think we need to change anything code-wise. Only the UI that the user interfaces with should not reflect it. Here, blacklist and whitelist signify what the preferences store.

akshaaatt: No @07jasjeet . We should avoid the usage like this here from the code as well.

app/src/main/java/org/listenbrainz/android/service/ListenScrobbleService.kt
07jasjeet: Called using lazy, should be same as calling in-place.

akshaaatt:
```kotlin
private val sessionManager: MediaSessionManager? by lazy {
    val manager = ContextCompat.getSystemService(this, MediaSessionManager::class.java)
    if (manager == null) {
        Log.e("MediaSessionManager is not available in this context.")
    }
    manager
}
```
something like this is good. Avoid `!!` as much as you can

build.gradle
```diff
@@ -17,7 +17,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.2.1'
+        classpath 'com.android.tools.build:gradle:8.1.4'
```
akshaaatt, line 20: This is wrong.

app/src/main/AndroidManifest.xml
```diff
@@ -57,7 +57,7 @@
             android:theme="@style/AppTheme" />
 
         <service
-            android:name=".service.ListenScrobbleService"
+            android:name=".service.ListeningService"
```
akshaaatt, line 60: This naming is wrong because it sounds very generic and is confusing. Something like `ListenSubmissionService` is better

app/src/main/java/org/listenbrainz/android/di/brainzplayer/DaoModule.kt
```diff
@@ -37,6 +37,6 @@ object DaoModule {
     
     @Provides
     fun providesPendingListensDao(
-        database: ListensScrobbleDatabase
+        database: ListensDatabase
```
akshaaatt, line 40: Same with this

akshaaatt (changes requested):

app/src/main/AndroidManifest.xml
07jasjeet: Sounds good. Will do.

build.gradle
07jasjeet: It's throwing incompatibility error for AS giraffe.

akshaaatt: Please upgrade to AS Hedgehog

akshaaatt: If not, please don't commit this. Revert for now

07jasjeet: Looks like its high time to switch

app/src/main/java/org/listenbrainz/android/di/brainzplayer/DatabaseModule.kt
```diff
@@ -25,11 +25,11 @@ object DatabaseModule {
...
     @Singleton
-    fun providesListensScrobbleDatabase(
+    fun providesListenSubmissionDatabase(
         @ApplicationContext context: Context
-    ): ListensScrobbleDatabase = Room.databaseBuilder(
+    ): ListensSubmissionDatabase = Room.databaseBuilder(
         context,
-        ListensScrobbleDatabase::class.java,
+        ListensSubmissionDatabase::class.java,
         "listens_scrobble_database"
```
akshaaatt, line 33: TODO: change this later

app/src/main/java/org/listenbrainz/android/repository/preferences/AppPreferencesImpl.kt
akshaaatt: We decided to change this later

app/src/main/java/org/listenbrainz/android/application/App.kt
akshaaatt: @07jasjeet have we fixed this?

akshaaatt: Please fix this @07jasjeet

07jasjeet: Its already fixed, this is an older commit.

akshaaatt: Cool

akshaaatt (approved):
LGTM. Thanks! 💯

PR #348, opened 2024-01-24
FIxed YIM CRASH after Playlist

pranavkonidena:
# Problem
YIM on android was crashing after the playlists screen

# Solution
I realized that the api structure has changed after hitting it once. So, I have changed the YimData structure a bit (basically their serialized json names) and have also added nullorempty checks for good measure.

# Action
Could someone please test locally and confirm

# Demo
Please check IRC, have posted there


akshaaatt (approved):
Tested on my device and all works fine. Great work!

PR #351, opened 2024-01-28
Concurrency and typo fixes

07jasjeet:
## Changelog
1) Concurreny fix in `SpotifyAppRemote` connection functions that caused a crash sometimes. 
    Why? Since Spotify sometimes runs callbacks twice, it makes our coroutine resume twice which causes an exception.
2) Typo fix.

Note: Checkout [this](https://github.com/metabrainz/listenbrainz-android/commit/f20f9f8bf93c76d0cfc44e6bdc0ea16735b06952) commit's description where concurrency fix is reverted.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -31,8 +31,8 @@ android {
         applicationId 'org.listenbrainz.android'
         minSdk 21
         targetSdk 34
-        versionCode 46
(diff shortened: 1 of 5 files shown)


akshaaatt (approved):
Interesting. Thanks!

Issue #330, opened 2024-01-08
kapt to ksp migration


utkarsh006:
Please report issues to the most appropriate project and component at
[Issue Link](https://tickets.metabrainz.org/browse/LB-1447)

For informal discussions, simply join MetaBrainz community forums at
https://community.metabrainz.org

For instant chat about development, join #metabrainz IRC channel at
ircs://irc.libera.chat:6697/metabrainz

akshaaatt:
Irrelevant for now

PR #357, opened 2024-02-05
kapt to ksp migration (Work in Progress)

utkarsh006:
- Solves #330


akshaaatt (commented):
Why was this required? Also, what are the benefits this brings? I could have just given this file to an LLM and asked it to convert for me.

utkarsh006:
@akshaaatt What's LLM? The issue is mentioned in PR and the purpose is mentioned [here](https://tickets.metabrainz.org/browse/MOBILE-183)

akshaaatt:
@utkarsh006 My statements were clear, and your comprehension deviates markedly from the intended message. We operate on a principle that requires logical reasoning behind every action. If you are keen on contributing, it is imperative that you first focus on the tickets we have prepared. I strongly advise initiating a conversation with us on IRC before taking any steps.

utkarsh006:
@akshaaatt Cool akshat, your statements were clear, so I deviated and yes I responded the same way, your tone of reply was to me. I would suggest you being simple with your words with the future contributors.

PR #366, opened 2024-02-15
MOBILE-188/Features such as recommend,review in profile screen

pranavkonidena:
# Link to ticket
https://tickets.metabrainz.org/browse/MOBILE-188?filter=-2

# Problem
The features in the feed such as writing review , recommend etc were not present in the profile screen even tho the listen cards were there. This made it very bland, hence I refactored so that those features can be done from the profile screen as well!

# Solution
I used the socialViewModel and called fns for pin and recommend to followers directly. For review and personal recommendation, I had to use feedViewModel as well for filtering out user's followers and had to make a dialog system as was done for the feed. Also , I made sure a toast (snackbar) is displayed every time a successful event is triggered so that the user gets feedback as well. I have added those messages inside constants file as they may be reused.

# Demo
https://github.com/metabrainz/listenbrainz-android/assets/122373207/0aaae0ed-a7ad-49be-a722-bfc3fcb420dc

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/listens/ListensScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/listens/ListensScreen.kt
@@ -61,6 +61,7 @@ import org.listenbrainz.android.viewmodel.SocialViewModel
 fun ListensScreen(
     viewModel: ListensViewModel = hiltViewModel(),
     socialViewModel: SocialViewModel = hiltViewModel(),
+    feedViewModel : FeedViewModel = hiltViewModel(),
     scrollRequestState: Boolean,
     onScrollToTop: (suspend () -> Unit) -> Unit,
     snackbarState : SnackbarHostState
@@ -69,15 +70,21 @@ fun ListensScreen(
     val uiState by viewModel.uiState.collectAsState()
     val preferencesUiState by viewModel.preferencesUiState.collectAsState()
     val socialUiState by socialViewModel.uiState.collectAsState()
-    
+    val feedUiState by feedViewModel.uiState.collectAsState()
+    val dropdownItemIndex: MutableState<Int?> = rememberSaveable {
+        mutableStateOf(null)
+    }
+
     ListensScreen(
         scrollRequestState = scrollRequestState,
         onScrollToTop = onScrollToTop,
         uiState = uiState,
+        feedUiState = feedUiState,
         preferencesUiState = preferencesUiState,
         updateNotificationServicePermissionStatus = {
             viewModel.updateNotificationServicePermissionStatus()
         },
+        dropdownItemIndex = dropdownItemIndex,
         validateUserToken = { token ->
             viewModel.validateUserToken(token)
         },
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/listens/ListensScreen.kt
```diff
@@ -145,12 +190,73 @@ fun ListensScreen(
...
+                    dropDown = {
+                        SocialDropdown(
+                            isExpanded = dropdownItemIndex.value == index,
+                            onDismiss = {
+                                dropdownItemIndex.value = null
+                            },
+                            metadata = metadata,
+                            onRecommend = {
+                                try {
+                                    socialViewModel.recommend(metadata)
```
07jasjeet, line 203: This is wrong, we do not propagate errors to the UI right now so this will always result in a successful toast. If you want, you can propagate the error via `Resource` class and then check if the response was successful then return error. We can ditch errorFlow for this as well.

app/src/main/java/org/listenbrainz/android/ui/screens/listens/ListensScreen.kt
```diff
@@ -164,6 +270,75 @@ fun ListensScreen(
...
+    onReview : (type: ReviewEntityType, blurbContent: String, rating: Int?, locale: String , metadata : Metadata) -> Unit,
+    onPersonallyRecommend : (metadata : Metadata , users : List<String> , blurbContent : String) -> Unit
+
+) {
+    val context = LocalContext.current
+    when (currentDialog) {
+        Dialog.NONE -> Unit
+        Dialog.PIN -> {
+            PinDialog(trackName = listens[currentIndex!!].trackMetadata.trackName, artistName = listens[currentIndex].trackMetadata.artistName, onDismiss = deactivateDialog, onSubmit = {
+                blurbContent -> try {
```
07jasjeet, line 292: Same here as well. Please take care everywhere you've done this.

app/src/main/java/org/listenbrainz/android/ui/screens/listens/ListensScreen.kt
07jasjeet: Remember its okay to change the core code if you feel its right. At most I will say is "This is wrong" and you will learn a thing or two. Great work overall.

07jasjeet (changes requested):

07jasjeet:
Also, can we please use Snackbar instead of toasts. I will choose the same on feed as well so it makes sense we have it here.

app/src/main/java/org/listenbrainz/android/ui/screens/listens/ListensScreen.kt
```diff
@@ -60,24 +84,50 @@ fun ListensScreen(
...
+}
+
+
 
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
 fun ListensScreen(
     scrollRequestState: Boolean,
     onScrollToTop: (suspend () -> Unit) -> Unit,
+    socialViewModel: SocialViewModel = hiltViewModel(),
```
07jasjeet, line 112: Please lift view-models out of this layer to the upper layer. This makes previews possible. Pass all the functions and states you need as parameters.

07jasjeet (changes requested):
Great work! After these last set of changes, I think we would be good to go :)

pranavkonidena:
@07jasjeet Please review the updated PR when u r able! Thanks :)

07jasjeet:
I think we can improve on the structure by just using one flow but LGTM for now. @akshaaatt good to merge.

app/src/main/java/org/listenbrainz/android/util/Constants.kt
```diff
@@ -35,6 +35,11 @@ object Constants {
         const val REFRESH_TOKEN = "refresh_token"
         const val STATUS_LOGGED_IN = 1
         const val STATUS_LOGGED_OUT = 0
+        const val RECOMMENDATION_GREETING = "Song recommended successfully!"
```
akshaaatt, line 38: @pranavkonidena Please move these to the strings file in res. That would enable us to be able to translate these.

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/util/Constants.kt
akshaaatt: @pranavkonidena please remove the strings from Constants and use them directly from res/strings

pranavkonidena: @akshaaatt So sorry , I misunderstood what I needed to do.. hopefully it is fixed now

PR #368, opened 2024-02-16
Quality of life fixes

07jasjeet:
# Change log
1) Feed horizontal scrolling should be smoother now.
2) When the server was down, we weren't able to parse the error correctly and show to the user, its fixed now.
3) Accompanist pager caused some issues with paging and was outdated in general. Used compose foundation pager instead.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/util/Log.kt
+++ b/app/src/main/java/org/listenbrainz/android/util/Log.kt
@@ -2,17 +2,24 @@ package org.listenbrainz.android.util
 
 import com.limurse.logger.Logger
 
-object Log {
-
-    fun e(message: Any?, tag: String? = null) {
-        Logger.e(tag, msg = message.toString())
-    }
-
-    fun d(message: Any?) {
-        Logger.d(msg = message.toString())
-    }
+interface Log {
+    fun e(message: Any?, tag: String? = null)
     
-    fun w(message: Any?) {
-        Logger.w(msg = message.toString())
+    fun d(message: Any?)
+    
+    fun w(message: Any?)
+    
+    companion object: Log {
+        override fun e(message: Any?, tag: String?) {
+            Logger.e(tag, msg = message.toString())
+        }
+        
+        override fun d(message: Any?) {
+            Logger.d(msg = message.toString())
+        }
+        
+        override fun w(message: Any?) {
+            Logger.w(msg = message.toString())
+        }
     }
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -225,9 +225,6 @@ dependencies {
     testImplementation project(path: ':sharedTest')
     androidTestImplementation project(path: ':sharedTest')
 
-    //ViewPager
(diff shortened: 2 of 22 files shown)


app/src/main/java/org/listenbrainz/android/util/Log.kt
```diff
@@ -2,17 +2,24 @@ package org.listenbrainz.android.util
...
-object Log {
-
-    fun e(message: Any?, tag: String? = null) {
-        Logger.e(tag, msg = message.toString())
-    }
-
-    fun d(message: Any?) {
-        Logger.d(msg = message.toString())
-    }
+interface Log {
```
akshaaatt, line 5: This is one of the funniest code implementations I have ever seen XD

I mean... there is no need of an interface here when we are gonna make the default implementation stay the same. The only reason I kept this file was because I did not want to refactor Log.d and stuff everywhere in the existing codebase. But I think we should directly use Logger.d , etc everywhere. No need to maintain this file

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/util/Log.kt
07jasjeet: Logger does file writes underneath and instantiates new thread. It makes the components un-testable when I directly use Logger.x, this interface is just hack to make things testable. Although now we have to pass log as a parameter but small giveaway for big gains.

akshaaatt: @07jasjeet the logger shouldn't be a part of the test suite anyway. How come the logger is affecting the tests? We mock files for testing and logger shouldn't be used there. The logger is only supposed to be used in our implementations

akshaaatt: Could you please mention the tests the logger directly being used is affecting?

07jasjeet: In the #371, I was working with JobQueue, and using logger, it just throws errors because we are using android level components in unit tests. Same with when I was implementing Timer's tests.

akshaaatt: @07jasjeet Makes sense when we need to test. You will have to inject the logger class independently. The other way could be to decouple the logger entirely and implement it early on. So for our use case, this is sadly not a hack but the best way to deal with this..

akshaaatt:
@07jasjeet this PR is incomplete and the build itself breaks. There have been classes used here that are unavailable in the code. This is an incorrect way to present a PR. We cannot merge this until it passes the CI

07jasjeet:
@akshaaatt i see that, can we close this one and deal with everything in #371?

akshaaatt:
@07jasjeet Sure, works for me

07jasjeet:
Closing, #371 will be center of attention for the mentioned issues now.

PR #371, opened 2024-02-21
Feat: Implemented Job Queue

07jasjeet:
## Change log
1) Feed horizontal scrolling should be smoother now.
2) When the server was down, we weren't able to parse the error correctly and show to the user, its fixed now.
3) Accompanist pager caused some issues with paging and was outdated in general. Used compose foundation pager instead.
4) Replaced handler with custom JobQueue based on coroutines.

## Why?
Handler produced inconsistent results, sometimes there was delay and sometimes actions did not even complete. This new JobQueue ensures consistency. Handler carried out operation on main thread, JobQueue carries it out on Worker threads with added optimisation of coroutines. Creating a new thread creates a lot of overhead and uses way more resources than coroutines do, hence this solution.


app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
```diff
@@ -0,0 +1,169 @@
...
+import kotlin.coroutines.CoroutineContext
+
+class JobQueue(
+    private val dispatcher: CoroutineDispatcher,
+    private val log: Log = Log.Companion
+) {
+    private val exceptionHandler = CoroutineExceptionHandler { _, e ->
+        log.e("JobQueue Exception: ${e.message}")
+    }
+    private val scope = CoroutineScope(SupervisorJob() + dispatcher)
```
akshaaatt, line 30: Add the exceptionHandler as well here

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
```diff
@@ -0,0 +1,169 @@
...
+    }
+    
+    fun cancel() {
+        queue.cancel()
+        scope.cancel()
+    }
+    
+    companion object {
+        /** A class containing reference to a scheduled job and its token.*/
+        private data class QueueJob(
```
akshaaatt, line 163: You can keep this in the model folder directly and no need to keep this inside the companion object

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
```diff
@@ -0,0 +1,169 @@
...
+    * a specified token. */
+    private val jobsMap = HashMap<Any?, LinkedList<QueueJob>>()
+    private val queue = Channel<QueueJob>(Channel.UNLIMITED)
+    
+    // Locks.
+    private val syncLock = Mutex()
+    private val mapLock = Mutex()
+    
+    init {
+        with(scope) {
```
akshaaatt, line 42: just do scope.launch here directly

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
```diff
@@ -0,0 +1,169 @@
...
+    }
+    
+    fun postDelayed(
+        delayMillis: Long,
+        token: Any? = null,
+        context: CoroutineContext = dispatcher,
+        block: suspend CoroutineScope.() -> Unit
+    ) {
+        scope.launch {
+            val queueJob = createQueueJob(token, context, block)
```
akshaaatt, line 91: This logic here is redundant. Just do delay first and then call post() directly (the method above)

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
```diff
@@ -0,0 +1,169 @@
...
+                // are for safety.
+                jobsMap[queueJob.token]?.add(queueJob)
+            } else {
+                jobsMap[queueJob.token] = LinkedList<QueueJob>().apply { add(queueJob) }
+            }
+        }
+    }
+    
+    private suspend fun sendJobToQueue(queueJob: QueueJob) {
+        syncLock.withLock {
```
akshaaatt, line 151: Remove as this is redundant. queue.send is already synchronous

akshaaatt (changes requested):

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
07jasjeet: Noice catch

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
07jasjeet: My first approach was this only, but we need to add the job to `jobsMap` before delay so that we can mark the job as cancelled even when the delay is taking place.

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
07jasjeet: This would pollute the namespace without any benefit. Since the use of `QueueJob` is inside this class only, lets keep it here.

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
07jasjeet: We don't want to handle exceptions in any other coroutines. I want the app to crash so we can identify any malfunctions (Since we don't have QA). 99% of the time, the there will be no effect.

akshaaatt: This is a really bad practice @07jasjeet . We cannot afford to crash the app in any situation. Tbh even if there is an error, don't crash. Rather send toasts or errors. A crash leads to a direct 1 star rating of the app. Also, facing a crash helps no one. It just is a very bad experience. Move this.

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
akshaaatt: Don't have strong opinions here so if you are so into it then we can keep it here

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
akshaaatt: @07jasjeet I see that you didn't remove suspend from this function even after the update. Also, since this a one-liner anyway, we can omit keeping this function and use the logic directly at places

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
akshaaatt: Cool, makes sense

app/src/main/java/org/listenbrainz/android/util/JobQueue.kt
07jasjeet: Well I would still want the current behaviour in beta but if you say so, Ill add the handler.

akshaaatt (approved):
Looks good for now. Let's get this tested!

PR #382, opened 2024-03-02
Feat: BP Revamp in LB-Android Part 1 (Completed Overview Screens)

pranavkonidena:
This PR aims to revamp the existing BP with the new figma and features as given on the [GSoC Ideas Page](https://wiki.musicbrainz.org/Development/Summer_of_Code/2024/ListenBrainz). The updated figma file can be found [here](https://www.figma.com/file/tFLTQiq7QlA1oGRTJVtXml/ListenBrainz-App?type=design&node-id=351%3A956&mode=design&t=Cauk64QOJAR9Ui2e-1)

PR #395, opened 2024-04-23
BP Revamp (Cleanup and Part-2)

pranavkonidena:
This PR is for the cleanup of code post the [first section of BP revamp](https://github.com/metabrainz/listenbrainz-android/pull/382) and implementing new screens..


akshaaatt (approved):
@pranavkonidena I think we should remove force bangs from everywhere to avoid crashes. Also, more cleanup needs to be done. Let's do that in the next PR whenever you can.

pranavkonidena:
Yup will do that

PR #469, opened 2024-08-12
Mobile 203/user pages 1.4(Unit Tests)

pranavkonidena:
This PR adds the repository, view model and Unit Tests to the User Pages

PR #475, opened 2024-08-18
Mobile 203/Artist Pages 2.1

pranavkonidena:
Please don't merge before #469 gets merged
This PR adds the artist pages


app/src/main/java/org/listenbrainz/android/di/ArtistRepositoryModule.kt
pranavkonidena, line 1: EOF

akshaaatt (approved):
This is quality work @pranavkonidena 💯 

Found a couple of UI bugs here and there but let's fix them at last in another PR.

PR #485, opened 2024-09-23
General app fixes

07jasjeet:
## Listen service fix
Two approaches:
1. Using Coroutines.
2. Using handler.

## UI Fixes
1. Taste section fixes in listen card where `""` was shown for no `blurbcontent` in pin feed card.


07jasjeet:
@akshaaatt, Going for a Beta Release. Current approach works fine so far even in deep sleep.

PR #505, opened 2024-12-12
Fixed Song Duration bug + Player Screen UI changes

harsh-bhadauria:
# Overview
The following changes have been made in this PR:

1. Fixed the song duration showing as 0:00 by passing duration in BrainzPlayerExtensions.kt
1. On the player screen, made a bunch of ui changes for consistency
   * The row containing the song title and the heart icon now uses SpaceBetween (for proper spacing)
   * Made the padding of song title row, progress bar and buttons row somewhat consistent
   * The cards below "Listening now" are no longer bound to max 600.dp (this caused issues when the phone was rotated, see image below) and are now properly spaced.

Please let me know if any further changes are required.

# Screenshots

Issue Screenshots:

WhatsApp Image 2024-12-12 at 12 28 19 PM

Fixed player screen:

(cards fill the screen correctly)

---
Issue on Bug Tracker: [MOBILE-206](https://tickets.metabrainz.org/browse/MOBILE-206)

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
@@ -149,7 +149,8 @@ fun BrainzPlayerBackDropScreen(
             val songList = brainzPlayerViewModel.mediaItem.collectAsState().value.data ?: listOf()
             SongViewPager(
                 modifier = Modifier.graphicsLayer {
-                    alpha = ( backdropScaffoldState.requireOffset() / (maxDelta - headerHeight.toPx()) )
+                    alpha =
+                        (backdropScaffoldState.requireOffset() / (maxDelta - headerHeight.toPx()))
                 },
                 songList = songList,
                 backdropScaffoldState = backdropScaffoldState,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
```diff
@@ -188,14 +189,19 @@ fun PlayerScreen(
             }
         }
         item {
-            Row(verticalAlignment = Alignment.CenterVertically) {
-                Column(modifier = Modifier.fillMaxWidth(0.8f)) {
+            Row(
+                verticalAlignment = Alignment.CenterVertically,
+                horizontalArrangement = Arrangement.SpaceBetween,
+                modifier = Modifier
+                    .padding(start = 25.dp, end = 25.dp)
```
07jasjeet, line 196: Please use `horizontal = ListenBrainzTheme.paddings.horizontal` as the argument here instead of `25.dp` so that we have consistent side padding across the app. Previously it was broken as well, so better fix it now while we are at it :)

Also, I do not recommend using `ListenBrainzTheme.paddings` everywhere. Sometimes, hardcoded paddings are the way, so that is that. Feel free to explore what paddings are there in the provider itself and think if any options feel like a fit, if not, hardcode :)

07jasjeet (changes requested):

07jasjeet:
Hi, thanks for opening this PR and taking up this long standing issue. Overall, things look great, just tiny nits and then we are good to go!

harsh-bhadauria:
Hi @07jasjeet, I tried to use `ListenBrainzTheme.paddings` and here's what it looks like (hardcoded one on the right for comparison)

  

Personally think that the right one looks cleaner and less cluttered. Would it be possible to add some additional values to `ListenBrainzTheme.paddings` for the player screen? If not, I'll try to make the best of what we got (with some harcoded values here and there).

07jasjeet:
I see your point. Can we keep the current padding for listen cards below and for the player name and progress, revert to your old paddings. Just need the listen card's to be in line within rest of the app

harsh-bhadauria:
Alright! Thank you for considering my request. The cards, checkboxes and the delete button are now properly aligned and use `ListenBrainzTheme.paddings.horizontal` (previously the checkboxes were not in line with the maxWidth card). Here's how it looks-





Please suggest any further changes that may be needed.

07jasjeet:
LGTM, great work!

07jasjeet (approved):

07jasjeet:
Thanks for the contribution @harsh-bhadauria! As a side note, you can chat with the whole team by going through this guide https://wiki.musicbrainz.org/Communication and setting up your favorable mode of communication.

PR #519, opened 2025-01-01
Fix: Delete song from BrainzPlayer playlist

Shreyassp002:
This PR resolves a bug where songs were not being deleted from the playlist in the BrainzPlayer screen. The following changes were made.

- Updated the logic in `BrainzPlayerBackDropScreen.kt` to properly remove the selected songs (checkedSongs) from the playlist.
- Ensured that the `currentPlayable`'s song list is mutable and modified correctly during the deletion process.

These updates ensure a smoother user experience when managing playlist in the BrainzPlayer feature.

**Before:**

https://github.com/user-attachments/assets/fe542c04-80c4-4e1d-a483-fa6fcb7af186

**Fixed:**

https://github.com/user-attachments/assets/bbb86c7d-c833-4b15-9f59-317ea720c893

Let me know if there are any changes needed!


07jasjeet (approved):
Lgtm. We also need to re-think this checked box design-wise because its very out of line wrt to the app.

PR #520, opened 2025-01-04
feat: Made the player screen background dynamic

GautamCoder4019k:
This pull request includes several changes to improve the appearance and functionality of the BrainzPlayer UI by introducing dynamic background colors and enhancing the TopBar component. The most important changes include adding dynamic background color functionality to the player screen, updating the TopBar to accept a customizable background color, and integrating the BrainzPlayerViewModel to manage these changes.

### Dynamic Background Color Functionality:

* Introduced dynamic background color and brush to the player screen and modified `AlbumArtViewPager` to use the dynamic background color. 

### TopBar Customization:

* Added a `backgroundColor` parameter to the `TopBar` component to allow customizable background colors.

### ViewModel Integration:

* Added methods to the `BrainzPlayerViewModel` to compute and store the dominant color from the album art. This color is then used as the dynamic background color for the player screen.

### MainActivity Updates:

*Integrated `BrainzPlayerViewModel` to manage the dynamic background color and updated the `Scaffold` in `MainActivity` to use this color.

These changes collectively enhance the user experience by providing a visually dynamic and customizable interface for the BrainzPlayer.

Screenshots

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
@@ -6,7 +6,6 @@ import androidx.compose.animation.core.animateDpAsState
 import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
-import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxWithConstraints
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
```diff
@@ -118,11 +121,21 @@ fun BrainzPlayerBackDropScreen(
         mutableFloatStateOf(0F)
     }
     val repeatMode by brainzPlayerViewModel.repeatMode.collectAsStateWithLifecycle()
+    val context = LocalContext.current
+    val defaultBackgroundColor = MaterialTheme.colorScheme.background
+    val isSystemInDarkTheme = isSystemInDarkTheme()
```
07jasjeet, line 126: We have a custom theme preference, which is different from what this function returns here. You can get the exact ui mode by using  `LocalUIMode.current`. There is also another function called `getOnScreenUiMode` or something similar that handles this case. Can you re-evaluate the usages of this variable and function as well?

app/src/main/java/org/listenbrainz/android/viewmodel/BrainzPlayerViewModel.kt
```diff
@@ -67,6 +75,8 @@ class BrainzPlayerViewModel @Inject constructor(
     val repeatMode = brainzPlayerServiceConnection.repeatModeState
     var isSearching by mutableStateOf(false)
 
+    var playerBackGroundColor by mutableStateOf(Color.White)
```
07jasjeet, line 78: The initial value of this color will probably be out of sync with the theme but i guess we can let that slide this time :)

07jasjeet (changes requested):

app/src/main/java/org/listenbrainz/android/viewmodel/BrainzPlayerViewModel.kt
GautamCoder4019k: i thought this case will not happen since when we are initializing the player screen at that time this state is also updating so this will not affect much. We can keep Color.Transparent as initial if white is a problem

07jasjeet: @GautamCoder4019k i think that should work but please test it first

GautamCoder8: i have tested and it was working fine

GautamCoder4019k: @07jasjeet i have checked and it was working fine now.

07jasjeet:
Hi @GautamCoder4019k, Thanks for changes. I took some liberty to improve the feature and fix a few bugs here and there. You can go through [this](https://github.com/metabrainz/listenbrainz-android/pull/520/commits/89a669f62ca02dcc8d1c3ebb7956ef7859df3d3d) commit for what has been changed. Overall this looks like a nice addition.

07jasjeet (approved):

GautamCoder4019k:
Hi @07jasjeet , Thanks for improving the commit, I learned some valuable approaches from your changes that I’ll use going forward.

PR #518, opened 2024-12-28
MOBILE-212: Highlight current playing song & dynamic PlayButton

Shreyassp002:
This PR focuses on highlighting the title of the currently playing song  and changing the icon in `BrainzPlayerListenCard`.

**Changes:**

- Added a visual highlight to the title of the song currently playing in the list.
- Added logic to match the `currentlyPlayingTitle` with the title passed to the `BrainzPlayerListenCard` composable. If they match, the title color will be changed; otherwise, it uses the default text color from the theme.
- Added animated Play/Pause Icon transition in `BrainzPlayerListenCard`
- Also made some formatting fixes for better code readability.
 

**Screenshots:**
<p align="center">
  Screenshot 1
  Screenshot 2
</p>

**Issue Link:** [MOBILE-212](https://tickets.metabrainz.org/projects/MOBILE/issues/MOBILE-212)

This is the current implementation of feature. I’d love to hear your feedback or suggestions for improvement.

PR #523, opened 2025-01-09
Feat: Dropdown songs card for AlbumsOverViewScreen & ArtistsOverviewScreen

Shreyassp002:
This PR introduces a dropdown feature in the `ArtistsOverviewScreen` and `AlbumsOverViewScreen` to enhance the user experience by allowing users to view songs in each artist and album.

**Changes:**

- Added a dropdown to display the list of songs for each artist and album.
- Utilized `AnimatedVisibility` to animate the dropdown with fade-in and fade-out effects, as well as vertical expansion and shrinkage.
- Used `ListenCardSmall` to display the corresponding songs within the dropdown.

Result:

https://github.com/user-attachments/assets/6712f3b2-0046-40f8-87e3-5a7d6c7ad13f

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerScreen.kt
@@ -301,10 +301,21 @@ fun BrainzPlayerHomeScreen(
                     )
                 },
                 onAddToNewPlaylist = {
-                    artist ->  
+                    artist ->
                 },
                 onAddToExistingPlaylist = {
-                    artist ->  
+                    artist ->
+                },
+                onSongClick = {
+                    song ->
+                    brainzPlayerViewModel.changePlayable(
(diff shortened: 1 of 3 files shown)


07jasjeet:
Hi @Shreyassp002, as much as like this, unfortunately we should navigate the user to artist screen where they can view all the songs. In BP, we do not have a _BP artist screen_ as of now, so I'm not sure how we should go about this. As far as clicking goes, we would want clicking to be reserved for playing the song only and not some other action :)

Shreyassp002:
Hey @07jasjeet ,
Thanks for the clarification! Since BP doesn’t currently have an artist screen, I wanted to ask:

Instead of the play button icon, should I add a dropdown icon that will display the songs of that artist or album? Or should we wait for the implementation of the dedicated BP artist screen before adding such functionality?

Also, clicking to play the song on the `BrainzPlayerListenCard` will be handled in PR #518.

07jasjeet:
Let me see if we can reuse the current artist screen. Until then, we can mark this PR as a draft.

PR #531, opened 2025-01-24
Fixed Follow Button Of CurrentUser

rahul31124:
This PR resolves two main issues in the Follow Card component:

**Incorrect Text Display:** The text displayed as "Followers" in both the "Followers" and "Following" sections, which caused confusion.

**Follow Button Visibility Issue:**
The Follow button appeared next to the user's own profile in the followers/followings list of another user
Changes Made:
Fixed Incorrect Text: The text now properly reflects whether the user is in the Followers or Following list.
The follow button will not appear next to the user's own profile in the followers/following list any other user.

**Before**
[Follow_Issue_Before.webm](https://github.com/user-attachments/assets/a2169a89-3eb2-4c29-b792-5d9b12bfa6e7)
**After**
[Follow_Issue_After.webm](https://github.com/user-attachments/assets/b5b3735e-f011-40aa-b46c-be6d4e2cedd4)

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/profile/listens/ListensScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/profile/listens/ListensScreen.kt
@@ -57,6 +57,8 @@ import androidx.compose.ui.text.withStyle
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
+import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import org.listenbrainz.android.R
 import org.listenbrainz.android.model.Listen
 import org.listenbrainz.android.model.Metadata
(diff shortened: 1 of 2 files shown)


07jasjeet (changes requested):
Hi @rahul31124, this task would require you to refactor every usage of follow button and hoist its state properly, please use view-models in high level components only. This for reasons already explained in review comment.

app/src/main/java/org/listenbrainz/android/ui/screens/profile/listens/ListensScreen.kt
```diff
@@ -838,9 +841,11 @@ private fun FollowersCard(
             }
         }
         Spacer(modifier = Modifier.height(10.dp))
+        val dashboardViewModel: DashBoardViewModel = hiltViewModel()
```
07jasjeet, line 844: Injecting view-models into low level components hinders our ability to test and render previews because previews cannot instantiate view-models. You would probably need to get this view-model from a high level component (its usage site) and then hoist the `isSelf` state from there.

07jasjeet (changes requested):
Hi @rahul31124, this task would require you to refactor every usage of follow button and hoist its state properly, please use view-models in high level components only. This for reasons already explained in review comment.

app/src/main/java/org/listenbrainz/android/ui/screens/profile/listens/ListensScreen.kt
rahul31124: @07jasjeet Hi Sir,I have implemented the State Hoist. Could you please check if everything looks good?

07jasjeet:
Hi, I've pushed some changes to make things correct. Please go through the changes to derive knowledge of the codebase. Primarily, I want to focus change where I removed the view-model being used in a layer which is preview-able. This is a behaviour we want to achieve **everytime** and build towards it.

PR #533, opened 2025-01-25
Bugfix: null song is coming when we click on recently played

GautamCoder4019k:
This PR fixes a bug where clicking on recently played list a null song gets added to the playlist. The issue occurred due to adding a empty song object to the recent played list and that empty song object was just used for checking if we should show all recently played card or not. In this pr i changed the logic for that and also refractor the code.

Earlier

https://github.com/user-attachments/assets/8cfae476-5d30-45ec-86ee-66689c7a70bf

Now

https://github.com/user-attachments/assets/6864c3d6-d3cc-4da8-95e7-33347dffd350


07jasjeet:
Thanks for the contribution @GautamCoder4019k!

PR #534, opened 2025-01-26
Fixed UI State Loss In BrainzPlayerSearchScreen 

rahul31124:
This PR resolves the issue of UI state loss in BrainzPlayerSearchScreen during screen orientation changes. Previously, the search query and UI state were reset when the screen rotated. This issue has been fixed by introducing state retention in BrainzPlayerViewModel and ensuring that the UI state, including the search query and results, is preserved across orientation changes.

**Before**

https://github.com/user-attachments/assets/791c61e9-cf2b-4ae1-a9d0-42021e313007

**After**

https://github.com/user-attachments/assets/89b85ead-3c7d-4a90-96e4-1f12da8c6632

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/search/BrainzPlayerSearchScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/search/BrainzPlayerSearchScreen.kt
@@ -78,7 +78,6 @@ fun BrainzPlayerSearchScreen(
             onQueryChange = { newValue: String ->
                 val updatedQuery = TextFieldValue(newValue, selection = brainzplayerQueryState.selection)
                 viewModel.updateSearchQuery(updatedQuery)
-                viewModel.searchSongs(newValue)
             },
             onClear = {
                 viewModel.clearSearchResults()
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/search/BrainzPlayerSearchScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/search/BrainzPlayerSearchScreen.kt
@@ -18,6 +18,7 @@ import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateListOf
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/listenbrainz/android/ui/screens/search/BrainzPlayerSearchScreen.kt
```diff
@@ -71,20 +67,22 @@ fun BrainzPlayerSearchScreen(
...
             },
             onDismiss = ::onDismiss,
-            onQueryChange = { newValue  ->
-                brainzplayerQueryState = newValue
-                searchItems.clear()
-                searchItems.addAll(viewModel.searchSongs(brainzplayerQueryState) ?: emptyList())
+            onQueryChange = { newValue: String ->
+                val updatedQuery = TextFieldValue(newValue, selection = brainzplayerQueryState.selection)
+                viewModel.updateSearchQuery(updatedQuery)
+                viewModel.searchSongs(newValue)
```
07jasjeet, line 81: Hi, we should not call this search function here itself, but in the view-model. We want to debounce this request by about 100-200 millisecs (since its local search, otherwise we prefer 500 when network calls are required). Please take inspiration from the following line: https://github.com/metabrainz/listenbrainz-android/blob/de22834b8c97c88862d9e25094060027fab38c37/app/src/main/java/org/listenbrainz/android/viewmodel/SearchViewModel.kt#L40

You can go through [SearchViewModel.kt](https://github.com/metabrainz/listenbrainz-android/blob/main/app/src/main/java/org/listenbrainz/android/viewmodel/SearchViewModel.kt) to derive your logic for searching.

07jasjeet (changes requested):
Hi @rahul31124, thanks for picking this up. You hit the correct spot by using view-model to preserve state!! But, there are better ways to do this task for which I have left comments in the review itself :)

app/src/main/java/org/listenbrainz/android/ui/screens/search/BrainzPlayerSearchScreen.kt
rahul31124: Hi @07jasjeet ,
I have refactored the BrainzPlayerViewModel and BrainzPlayerSearchScreen as per the guidance,could you please review and let me know if everything looks good?
Thank you!

rahul31124: @07jasjeet ?

07jasjeet: Looks good!

07jasjeet (approved):
LGTM, Great work!

PR #548, opened 2025-02-03
feat: made the ui adaptive for orientation changes

GautamCoder4019k:
in this pr the ui for the app is changed so that it is automatically handles orientation changes and ui is more responsive

before

https://github.com/user-attachments/assets/4e639811-c8b0-4d6b-b4f9-cdd5f48d0341

after

https://github.com/user-attachments/assets/3351b2fd-07f5-47e7-ad99-90055ee7253b


app/src/main/java/org/listenbrainz/android/ui/navigation/SideNavigationBar.kt
```diff
@@ -0,0 +1,300 @@
...
+        Spacer(modifier = Modifier.weight(1f))
+        BrainzNavigationMiniPlayer(
+            backdropScaffoldState = backdropScaffoldState,
+            currentlyPlayingSong = currentlyPlayingSong,
+        )
+    }
+}
+
+@Composable
+fun BrainzNavigationMiniPlayer(
```
07jasjeet, line 180: We need to re-use existing component somehow by making it flexible to resizing. Maintaining 2 components is a pain when it comes to updating something.

app/src/main/java/org/listenbrainz/android/ui/navigation/SideNavigationBar.kt
```diff
@@ -0,0 +1,300 @@
...
+import org.listenbrainz.android.model.Song
+import org.listenbrainz.android.ui.components.CustomSeekBar
+import org.listenbrainz.android.ui.components.PlayPauseIcon
+import org.listenbrainz.android.ui.screens.brainzplayer.ui.components.basicMarquee
+import org.listenbrainz.android.ui.theme.ListenBrainzTheme
+import org.listenbrainz.android.util.BrainzPlayerExtensions.toSong
+import org.listenbrainz.android.viewmodel.BrainzPlayerViewModel
+
+@Composable
+fun SideNavigationBar(
```
07jasjeet, line 67: Same with this code, most logic remains same but a new component will make maintenance difficult.

07jasjeet (changes requested):

app/src/main/java/org/listenbrainz/android/ui/navigation/AdaptiveNavigationBar.kt
```diff
@@ -0,0 +1,236 @@
...
+@Composable
+fun AdaptiveNavigationBar(
+    modifier: Modifier = Modifier,
+    navController: NavController = rememberNavController(),
+    backgroundColor: Color = ListenBrainzTheme.colorScheme.nav,
+    backdropScaffoldState: BackdropScaffoldState = rememberBackdropScaffoldState(initialValue = BackdropValue.Revealed),
+    scrollToTop: () -> Unit,
+    username: String?,
+    isLandscape: Boolean,
+    brainzPlayerViewModel: BrainzPlayerViewModel = hiltViewModel()
```
07jasjeet, line 60: We need lift this view-model out of this component so as to make previews possible.

07jasjeet (changes requested):
Good stuff!, exactly like I wanted it to be! One small nit though and then I think we can move forward.

app/src/main/java/org/listenbrainz/android/ui/navigation/AdaptiveNavigationBar.kt
GautamCoder4019k: sorry was sick so was not able to work on it

07jasjeet (approved):
Lgtm, great work. This will definitely improve the experience on tablets!

PR #553, opened 2025-03-01
Fix null profile screen if we click on "You" in feed

GautamCoder4019k:
This pr is for the issue in feed  screen where  when a user taps on the "You" in the feed, the app navigates to the profile screen. However, the profile screen shows null values,

Before

https://github.com/user-attachments/assets/d4d86e84-52d1-4a58-b184-52665bc5aa3f

After

https://github.com/user-attachments/assets/2ddb6a4d-c15b-401b-8c24-62f9219afa36

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/model/feed/FeedEventType.kt
+++ b/app/src/main/java/org/listenbrainz/android/model/feed/FeedEventType.kt
@@ -1,6 +1,5 @@
 package org.listenbrainz.android.model.feed
 
-import android.content.ActivityNotFoundException
 import androidx.annotation.DrawableRes
 import androidx.compose.foundation.text.ClickableText
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/listenbrainz/android/ui/navigation/AppNavigation.kt
```diff
@@ -44,6 +44,7 @@ fun AppNavigation(
     scrollRequestState: Boolean,
     onScrollToTop: (suspend () -> Unit) -> Unit,
     snackbarState: SnackbarHostState,
+    user: String
```
07jasjeet, line 47: can you rename this to `loggedInUser` or something like that. 100 user/username variables confuse the logged in user and the selected user for which we are explicitly viewing profile of.

07jasjeet (changes requested):

07jasjeet:
Hi, on another look, I found out that getting "You" as annotation is not the expected behaviour I would want when "You" is clicked. Clicking on "You" should return the logged-in user's name itself.

I went on to find the root cause of this issue and its because it should be:
```kotlin
val annotation = feedEvent.metadata.user0 ?: feedEvent.username
addStringAnnotation(
    tag = "user0",
    annotation = annotation,
    start = 0,
    end = firstUsername.lastIndex + 1
)
```
in `FeedEventType.kt`. 

Please apply the changes and revert the "You" check in `goToArtistPage` function.

GautamCoder4019k:
Thanks for pointing this out, @07jasjeet! That makes sense. I'll apply the changes as suggested and revert the "You" check in goToArtistPage.

PR #555, opened 2025-03-06
Swipe Issue in landscape mode

rahul31124:
This PR fixes an issue in BrainzPlayerBackDropScreen where in landscape mode there was nothing to pull/swipe up it also refines alpha calculation in SongViewPager to prevent division by zero and applies safe constraints using .coerceAtLeast(1f) to avoid negative values

**Before**

https://github.com/user-attachments/assets/b9f03c24-317f-4e9a-9572-bde5583a232c

**After**

https://github.com/user-attachments/assets/04ed25ab-e2a5-49f4-a8a8-99b045ce579b

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/brainzplayer/BrainzPlayerBackDropScreen.kt
@@ -139,7 +139,7 @@ fun BrainzPlayerBackDropScreen(
 
     /** 56.dp is default bottom navigation height */
     val headerHeight by animateDpAsState(
-        targetValue = if (isLandscape) 0.dp else
+        targetValue = if (isLandscape) 28.dp else
             if (isNothingPlaying)
                 56.dp
             else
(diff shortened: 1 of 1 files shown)


GautamCoder4019k:
@rahul31124 There is already an option in landscape mode where clicking on the mini player in the navbar opens the player screen. So, this implementation might be redundant.

Additionally, in your approach, if the song has an image, the color from the image will be displayed at the bottom of the screen. This might not look visually appealing. @07jasjeet may have more insights on this.

07jasjeet:
Hi @rahul31124, I agree with @GautamCoder4019k on this. I specifically removed that bottom peek of the front layer sometime back because its not our intended UX. I appreciate your efforts, but you may need to close this PR. We have created some new issues on issue tracker, you can check them out :)

rahul31124:
Hi @07jasjeet,yeah clicking on the mini player opens the player screen but it can be confusing for users since there's no to swipe up. Initially a user has to swipe down to minimize, so they might naturally expect to swipe up to expand. I just realized now that clicking on the player screen expands it so no problem if you feel the current approach is not good. 👍 I will close this PR.

PR #556, opened 2025-03-18
MOBILE-214 Playlist-Detail-Screen & Add-Track-To-Playlist-Sheet

hemang-mishra:
I have implemented Playlist Detail Screen and add track to Playlist screen.
I have implemented features like:
1. Reorder items
2. Add tracks to playlist
3. Edit Playlist
4. Duplicate playlist

This is the last PR in the series. Playlist feature is completed.
I would be pleased to hear any feedback and modifications required.

## Screenshots for reference:
image
image


07jasjeet:
Hi, Great work! Some pointers:
1. We should show a dialog to confirm whether the user wants to duplicate a playlist before hitting the API.
2. On hitting back, I go back to feed and not profile. It should take me back to profile.
3. On search song sheet, we should add some space at end of the lazy list just as UX confirmation that there are no songs.
4. The search song sheet loader should be slightly padded from above.
5. On clicking `x`, the keyboard does not automatically show up and search bar is not focused.
6. On first open, keyboard should automatically open up and search bar should be focused so user can start typing instantly.

These are little UX details devs miss a lot of the times. Even experienced ones. Having a better knowledge of these UX quirks really defines the quality of work one does.

hemang-mishra:
@07jasjeet Thanks for the amazing feedback. I am trying my best to not miss these small UX details😅. 
I have fixed all the bugs except the 2nd. I am unable to replicate this issue in my device. Ideally it should go back to Profile (similar to Artist Screen as both have same implementation for navigation).

app/src/main/java/org/listenbrainz/android/ui/navigation/AppNavigation.kt
```diff
@@ -67,6 +70,12 @@ fun AppNavigation(
         }
     }
 
+    fun goToPlaylist(mbid: String){
+        navController.navigate("${AppNavigationItem.PlaylistScreen.route}/${mbid}"){
+            defaultNavOptions()
```
07jasjeet, line 75: This is the reason the app goes to feed when back is pressed, please rectify. Also, do the same for album page and artist page as well.

07jasjeet (changes requested):

app/src/main/java/org/listenbrainz/android/ui/navigation/AppNavigation.kt
hemang-mishra: I have updated the defaultNavOptions, hope this should resolve the issue.

app/src/main/java/org/listenbrainz/android/ui/navigation/AppNavigation.kt
```diff
@@ -44,10 +44,6 @@ fun AppNavigation(
     snackbarState: SnackbarHostState,
 ) {
     fun NavOptionsBuilder.defaultNavOptions() {
-        // Avoid building large backstack
-        popUpTo(AppNavigationItem.Feed.route) {
```
07jasjeet, line 45: We want this option to be here, rather hardcode options into each use case (artist, album and playlist) without this field.

07jasjeet (changes requested):

hemang-mishra:
@07jasjeet I want to just confirm the exact expectation before changing the navigation:
Currently, `popUpTo(AppNavigationItem.Feed.route)` is causing Profile to be removed from the backstack when navigating to Playlist. Should Profile always remain in the backstack when going to Playlist? Also, should this apply to other screens like Artist and Album? Or do we want a more general rule where back navigation always goes to the last visited screen?

07jasjeet:
> @07jasjeet I want to just confirm the exact expectation before changing the navigation: Currently, `popUpTo(AppNavigationItem.Feed.route)` is causing Profile to be removed from the backstack when navigating to Playlist. Should Profile always remain in the backstack when going to Playlist? Also, should this apply to other screens like Artist and Album? Or do we want a more general rule where back navigation always goes to the last visited screen?

First logic for Playlist, Artist and album. Rest should remain same. No common rule.

hemang-mishra:
@07jasjeet I have updated hardcoded the options for artist, playlist and album screen. Rests screens would be able to use defaultNavOptions

sharedTest/src/main/java/org/listenbrainz/sharedtest/mocks/MockPlaylistDataRepository.kt
```diff
@@ -27,15 +32,68 @@ class MockPlaylistDataRepository : PlaylistDataRepository {
...
-    ): Resource<String?> {
+    override suspend fun getPlaylistCoverArt(playlistMBID: String, layout: Int): Resource<String?> {
         return Resource(
             Resource.Status.SUCCESS,
             samplePlaylistCoverArt
         )
     }
 
+    override suspend fun addPlaylist(playlistPayload: PlaylistPayload): Resource<AddCopyPlaylistResponse?> {
+        return Resource(
```
07jasjeet, line 43: Whats going on here?

app/src/main/java/org/listenbrainz/android/ui/theme/Theme.kt
```diff
@@ -129,6 +130,14 @@ private val colorSchemeDark = ColorScheme(
             Color.Transparent
         )
     ),
+    playlistScreenGradient = Brush.verticalGradient(
```
07jasjeet, line 133: Is this the gradient aerozol was talking about? If yes, maybe name it to something else common since playlist is not the only thing where this is being used. Lets only keep common things here and not be this specific.

app/src/main/java/org/listenbrainz/android/repository/playlists/PlaylistDataRepositoryImpl.kt
```diff
@@ -34,26 +41,29 @@ class PlaylistDataRepositoryImpl @Inject constructor(
 
     override suspend fun getPlaylistCoverArt(
         playlistMBID: String,
-        dimension: Int,
         layout: Int
     ): Resource<String?> {
```
07jasjeet, line 45: Lets add `withContext(ioDispatcher)` on every function here. I figured its just very easy to re-use.

07jasjeet (changes requested):

app/src/main/java/org/listenbrainz/android/repository/playlists/PlaylistDataRepositoryImpl.kt
hemang-mishra: Done

sharedTest/src/main/java/org/listenbrainz/sharedtest/mocks/MockPlaylistDataRepository.kt
07jasjeet: @hemang-mishra ^^

app/src/main/java/org/listenbrainz/android/ui/theme/Theme.kt
07jasjeet: @hemang-mishra ^^

sharedTest/src/main/java/org/listenbrainz/sharedtest/mocks/MockPlaylistDataRepository.kt
hemang-mishra: Oops, sorry. I missed the other two comments. Saw the last one only.

hemang-mishra:
I have updated the test cases with a real MBID for test data of AddPlaylistResponse.
I followed the same way of testing like it was there in MockUserPlaylistRepository.

app/src/main/java/org/listenbrainz/android/ui/theme/Theme.kt
hemang-mishra: I was confused abt this. I asked about this once. I'll confirm it once from aerozol asap.

07jasjeet: Alright let it be like this right now, we can confirm later.

07jasjeet (approved):

PR #569, opened 2025-06-23
Feat: Listening apps implementation

hemang-mishra:
This PR implements Listening App Selection Screen.

Note: This PR is built on top of #567. Please review commits related to this feature only.

Method of fetching music apps:
1. First fetch of all audio related apps which have these intents:
`val intents = listOf(
                "android.media.browse.MediaBrowserService",
                "android.media.session.MediaSessionService"
            )`

2. After the apps are fetched, we still have some apps which are not adequate like Google and Bluetooth, we can filter furthur using categories:
`val category = pm.getApplicationInfo(packageName, 0).category
                    if(category == ApplicationInfo.CATEGORY_AUDIO || category == ApplicationInfo.CATEGORY_VIDEO){
                        musicAppsPackageNames.add(packageName)
                    }`
There is no specifc category for music, so I fetched these.

Let me know if I should separate youtube for these music apps and maybe keep it in a separate section.

### Screenshots:
image


07jasjeet:
3 issues:

1. Content is not edge to edge.
2. Toggle button is getting clipped.
3. Back button is not legible.

07jasjeet:
> @07jasjeet from ux pov, should we keep the back button scrollable with the rest of the content or keep it static with some padding like this.
> 
> image

Lets keep it floating and add a background to the button. The button takes too much space for no reason.

app/src/main/java/org/listenbrainz/android/viewmodel/DashBoardViewModel.kt
```diff
@@ -94,24 +98,147 @@ class DashBoardViewModel @Inject constructor(
...
+                appPreferences.requestedPermissionsList.getFlow().firstOrNull()?.toMutableList()
             if (permissions != null && !permissions.contains(permission.permission)) {
                 permissions.add(permission.permission)
                 appPreferences.requestedPermissionsList.set(permissions)
             }
         }
     }
 
-    fun markOnboardingComplete(){
+    fun getListeningApps(context: Activity) {
```
07jasjeet, line 110: A few things here:

1. This function can be separated into a util function and function which updates state inside this view-model. The util  function can be placed in the utils file.

2. This function is named `getListeningApps` but it does not return anything. Nomenclature is important in a codebase. Please name it `updateListeningApps` to avoid confusion.

3. The new util function can be an extension function with a return type as follows:
`Context.getListenableApps(): <List<ApplicationInfo>>`

07jasjeet (changes requested):

PR #571, opened 2025-07-20
make the full description a little more detailed

IzzySoft:
This PR adds some details to the EN full description, making it clearer what the app is about. It also slightly adjusts formatting, making it possible to render it as Markdown (supported e.g. at IzzyOnDroid) while keeping it compatible with places not supporting it (e.g. F-Droid.org, PlayStore).


07jasjeet:
Thanks for contribution!

IzzySoft:
Thanks for merging! :hugs:

PR #576, opened 2025-08-24
Feat: App Updates

hemang-mishra:
Implementation details for Github Updates:
1. Add a launch count, which counts the number of app launches which is critical for this implementation.
2. Added two app preferences, `lastVersionCheckLaunchCount` and `lastUpdatePromptLaunchCount` which respectively store when the was the latest version checked and when the last update was prompted to the user.
The duration of recheck and re-prompt are handled by these constants.
```
const val VERSION_CHECK_DURATION = 3
const val RE_PROMPT_USER_AFTER_DENIAL = 10
```
3. Updates are fetched using the github api and are compared with the current version of the build.
4. Separate handling of preview releases is also available.
5. Check for updates option is also added in Settings.
6. The updates are downloaded using Download Manager and the received id is stored in app preferences.
7. Broadcast receiver are setup whenever the updates are being downloaded to show install dialog whenever app update is downloaded.
8. Installing app updates requires a special permission and proper rationale is shown using a dialog.


07jasjeet (approved):
LGTM, super work. Refactored some things but overall solid work.

PR #580, opened 2025-09-08
Login improvement

hemang-mishra:
In this PR, 
1. Increased timeout to 1 min for slower internet connections.
2. Added an additional check if value returned by web view client in case of error is string "null", and giving proper rationale instead.
3. Added a "Cancel button" during the loading process so that user can stop the login process if they realize that wrong credentials were entered.


07jasjeet (approved):
LGTM nice work!

PR #592, opened 2025-10-17
Feat: Implement Sentry opt-in toggle

kartikey004:
- Added observer in DashBoardViewModel to listen to `isCrashReportingEnabled` preference
- Updated permission handling and onboarding flow to integrate crash reporting toggle
- Ensures Sentry can be enabled/disabled without restarting the app


app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/permissions/PermissionEnum.kt
```diff
@@ -125,6 +136,18 @@ enum class PermissionEnum(
                 val powerManager = context.getSystemService(Context.POWER_SERVICE) as PowerManager
                 powerManager.isIgnoringBatteryOptimizations(context.packageName)
             }
+            CRASH_REPORTING -> {
+                // Blocking read from DataStorePreference<Boolean> for onboarding checks
+                val pref = AppPreferencesImpl(context).isCrashReportingEnabled
```
07jasjeet: We need to use dependency injection rather than creating a new object here.

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/permissions/PermissionEnum.kt
```diff
@@ -125,6 +136,18 @@ enum class PermissionEnum(
...
+                var result = true
+                runCatching {
+                    kotlinx.coroutines.runBlocking {
```
07jasjeet: Think of ways we can get rid of runBlocking. Maybe make the function suspend?

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/permissions/PermissionEnum.kt
```diff
@@ -165,8 +190,8 @@ enum class PermissionEnum(
         }
         fun getPermissionsForPermissionScreen(): List<PermissionEnum> = PermissionEnum.entries.filter {
             it.isPermissionApplicable() &&
-            it != READ_NOTIFICATIONS && // READ_NOTIFICATIONS is handled separately
-            it != BATTERY_OPTIMIZATION // BATTERY_OPTIMIZATION is handled separately
+                    it != READ_NOTIFICATIONS && // READ_NOTIFICATIONS is handled separately
```
07jasjeet: I would recommend keeping spacing changes of untouched code out of a PR.

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/permissions/PermissionScreens.kt
```diff
@@ -236,6 +258,37 @@ fun PermissionCard(
     }
 }
 
+@Composable
+fun CrashReportingToggle(viewModel: DashBoardViewModel = hiltViewModel()) {
```
07jasjeet: Please re-use existing components and do not inject view-model into preview able composable functions. When you inject a hilt view-model (dependency injected class) into a composable function, the preview window is unable to create instance of the view-model because some classes are too complex to be instantiated like DBs or contexts and builds the preview in an isolated sandbox. You can however instantiate view-models in cases where you do not want previews.

app/src/main/java/org/listenbrainz/android/viewmodel/DashBoardViewModel.kt
```diff
@@ -71,6 +72,22 @@ class DashBoardViewModel @Inject constructor(
         }
     }
 
+    object CrashReporter {
```
07jasjeet: Here's how sentry's functions should be used in this class:

1. Create an interface named `SentryController` in another file.
2. Implement the above interface in that file itself and inject it via hilt into this view-model.
3. Use the injected instance to enable or disable sentry sending events.

kartikey004:
Got it, thanks for the review! I will make the updates soon.

07jasjeet:
Sentry has been removed from main project.

PR #618, opened 2025-11-17
Fixed bug of the bottom nav bar

### LINKED ISSUE #617: Fix Bottom Navigation Bar issue
There’s a bug in the bottom-nav flow.

From the Feed screen, if we go to Settings and then tap the Feed tab again, it should pop back to Feed, but right now it just stays stuck on Settings.

Same thing happens with the Profile tab.

Basically, whenever we navigate to any secondary screen (Settings, Album, Artist, Playlist, etc.) and then tap a bottom-nav tab, the current screen should be popped off the back stack and we should land on the selected tab’s root screen.

nirvan73:
FIXES #617 

This PR fixes the bug in the bottom navigation bar.
I checked the current route using the NavController, and if it is the Settings screen, I popped it before navigating to the desired screen from the bottom navigation bar.


07jasjeet (approved):

07jasjeet:
Thanks for the contribution @nirvan73!

PR #627, opened 2025-11-23
GDPR & LB Consent integration in Login flow

hemang-mishra:
Primary aim of this PR is to integrate these missing flows in the login process (which is only meant for new accounts):
1. Link MusicBrainz account with ListenBrainz.
2. GDPR compliance page.

Now, they are implemented by formatting the webpage using JS and presenting it in a box. The UI still looks clean because of formatting. Also implemented 4 listeners for events like accepting and rejecting the above mentioned compliance.

## Demo
Checkout the complete account creation flow:

https://github.com/user-attachments/assets/68abf761-c86f-46f7-a05e-789dfcfe7b76


07jasjeet (approved):
Works great for the most part as far as I've tested. Great work!

Issue #595, opened 2025-10-18
Navigation options for user


07jasjeet:
Currently, LB Android app (with a logged-in user) opens up Feed Section as the first screen. Some users prefer Profile as their first screen which means they have to click twice in-order for them to begin loading profile page. We want the user to have the option to set their navigation tabs  order such that the first tab will open up first.

Points to remember:
1. Allow users to set order of bottom navigation tabs via settings.
2. A standard setting option should open up another screen which displays bottom nav bar.
3. Holding and dragging any nav bar item should change its position and changes should be reflected in bottom bar immediately.
4. Datastore preference should be used to store the order, no BE involved.

AdvityaCode:
Hi @07jasjeet I'm not sure if @anuj99 got to finishing this issue, but I would like to take this up. Can it please be assigned to me?

rajeev-k47:
Hi @07jasjeet, I can take this up.

PR #635, opened 2025-12-19
Implemented the feature that allows reordering of Bottom navigation tabs

rajeev-k47:
**Implementation** for #595 

This PR adds support for customizable bottom navigation order. Users can now reorder bottom navigation tabs from Settings, and the first tab in the chosen order becomes the default landing screen on app launch. The order is stored locally using DataStore.

### Implementation recording:

https://github.com/user-attachments/assets/98a4bbf5-6de7-4080-9ceb-bbe87ab8299e


rajeev-k47:
@07jasjeet Ok i will do the changes, i have a doubt that can i change the click behavior of real navbar while reordering instead of covering it with a dummy navbar. Like the real navbar will then allow dragging of items making the default click events non functional.

Issue #615, opened 2025-11-16
Add Delete Listens


hemang-mishra:
We need to implement option to delete listens in dropdown below:
<img width="200" height="500" alt="Image" src="https://github.com/user-attachments/assets/fd605c1a-6387-44b3-a19f-d4b70b05057e" />

The endpoint for deleting listens can be found here: [Delete-listens](https://listenbrainz.readthedocs.io/en/latest/users/api/core.html#post--1-delete-listen)

Callbacks for `deleteListen` already exists in `SocialDropdown`, we just need to reuse it.

Since the listens are not deleted immediately after hitting the endpoint, remove the particular listen from the list on the screen after OK response.

07jasjeet:
@revaarathore11 we will assign this issue to @dibyajyoti-mandal given @dibyajyoti-mandal is still interested.

anuj990:
@dibyajyoti-mandal are you currently working on this issue, if not
@07jasjeet i would like to work on this issue

anuj990:
Hi @07jasjeet , following up on this issue Since @dibyajyoti-mandal hasn’t responded within the given timeframe, I’m available to take this up. Please assign the issue to me so I can start working on it

PR #636, opened 2025-12-21
Added delete listen feature

dibyajyoti-mandal:
implemented option to delete listens in dropdown #615

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/model/feed/FeedEventType.kt
+++ b/app/src/main/java/org/listenbrainz/android/model/feed/FeedEventType.kt
@@ -13,7 +13,9 @@ import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.withStyle
+import androidx.hilt.navigation.compose.hiltViewModel
 import org.listenbrainz.android.R
+import org.listenbrainz.android.model.Listen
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/listenbrainz/android/model/feed/FeedEventType.kt
```diff
@@ -183,22 +186,38 @@ enum class FeedEventType (
...
-                onOpenInMusicBrainz = onOpenInMusicBrainz,
-                onPin = onPin,
-                onReview = onReview,
-                onPersonallyRecommend = onPersonallyRecommend,
-                onRecommend = onRecommend,
-                goToUserPage = goToUserPage,
-                goToArtistPage = goToArtistPage
-            )
+            LISTEN -> {
+                val listensViewModel: ListensViewModel = hiltViewModel()
```
07jasjeet: Rather than creating a new view-model instance here, `onDeleteListen` should be passed in as a callback.

07jasjeet (changes requested):
Before reviewing further, I see that the project does not compile. Please make sure that the project compiles before requesting review. This happens to everyone now and then, but please make sure you sanity check your code before review.

As for the code itself, Please add delete listens function to `SocialViewModel` which can then be used to supply onDelete to `SocialDropdownDefault` which helps reduce the boilerplate of passing lambdas down the function hierarchy.

Also, before triggering delete, there should be certain checks included inside social dropdown such that if the given metadata is not suitable for delete operation, we should not show that option. This is already being done for other events as well.

PR #641, opened 2025-12-29
Fix playlist create/edit behavior 

### LINKED ISSUE #638: Edit Playlist doesn’t work on Playlist Tab under Profile Screen
### Steps to Reproduce:-

1. Open the ListenBrainz App
2. Navigate to the **Playlist** tab under the **Profile** screen
3. Tap on the three dots to open the `PlaylistDropDownMenu` and select the **Edit** option
4. Same issue occurs when you tap on a particular playlist
5. On the `PlaylistDetailScreen`, tap the **Edit playlist** button

### Expected Behavior:-

The Edit option should open the Modal Bottom Sheet with the **Edit State** and the initial data prefilled.

### Actual Behavior:-

The Edit option is opening the Modal Bottom Sheet with the **Create State** for a new playlist.

### Environment:-

- Android Version: 15
- App Version: 2.12.0.0
- Device: OnePlus Nord CE3

### Screen Recording: 

https://github.com/user-attachments/assets/1a758212-e3ac-479d-a24a-a1c2bd7dea51

nirvan73:
Fixes #638 

This PR fixes the bug in the create/edit behavior of a playlist in the **Playlist** tab under the **Profile** screen.

I moved the `LaunchedEffect` which contains `getInitialDataInCreatePlaylistScreen`, inside the `CreateEditPlaylistScreen` within the pre-existing if/else guard used to render the composable. As a result, the `LaunchedEffect`, even with Unit as its key, now runs only once when `isVisible` is toggled on. Hence resulting in proper flow of states of the Modal Bottom Sheet i.e:- Create State for creating a new playlist and Edit State for editing a particular playlist.

Screen Recording:-

https://github.com/user-attachments/assets/ea3ade97-4000-4400-82ce-aee0d817ad9a

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/playlist/AddEditPlaylistScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/playlist/AddEditPlaylistScreen.kt
@@ -94,15 +94,14 @@ fun CreateEditPlaylistScreen(
         }
     }
 
-    LaunchedEffect(Unit) {
-        viewModel.getInitialDataInCreatePlaylistScreen(mbid)
-    }
-
     LaunchedEffect(uiState.createEditScreenUIState.isSearching) {
         sheetState.expand()
     }
 
-    if (isVisible)
+    if (isVisible) {
+        LaunchedEffect(Unit) {
+            viewModel.getInitialDataInCreatePlaylistScreen(mbid)
+        }
         Box(
             modifier = Modifier.fillMaxSize()
         ) {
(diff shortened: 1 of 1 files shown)


hemang-mishra:
Good explanation here
https://github.com/metabrainz/listenbrainz-android/issues/638#issuecomment-3691631019
Apparently this issue occurred when we added the add to playlist option in the social dropdown, the guard fixes it effectively. Looks good!!

07jasjeet:
Thanks for the contribution!

PR #645, opened 2026-01-05
Feature addition [Reordering bottom nav bar]

### LINKED ISSUE #595: Navigation options for user

rajeev-k47:
Fixes #595 

PR continuing #635 

Hi @07jasjeet I have implemented the required features suggested by you. Kindly review it.
I used external library for implementing the drag and drop feature to avoid writing complex gesture handling and reordering animation logic manually. [Reorderable](https://github.com/Calvin-LL/Reorderable) supports compose multi-platform which will not create any issues while migrating to cmp. Current implementation of drag and drop in some components can also be replaced by this dep. Tell me if it looks fine to you.

https://github.com/user-attachments/assets/cc753a4c-2878-45a9-8a68-25449c83635d

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/main/HomeScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/main/HomeScreen.kt
@@ -261,7 +261,7 @@ fun HomeScreen(
                 deactivate = searchBarState::deactivate,
                 goToUserPage = { username ->
                     searchBarState.deactivate()
-                    navController.navigate("${AppNavigationItem.Profile.route}/$username")
+                    navController.navigate(AppNavigationItem.Profile.withUserArg(username))
                 }
             )
         }
--- a/app/src/main/java/org/listenbrainz/android/model/AppNavigationItem.kt
+++ b/app/src/main/java/org/listenbrainz/android/model/AppNavigationItem.kt
@@ -6,7 +6,13 @@ import org.listenbrainz.android.R
 sealed class AppNavigationItem(val route: String, @DrawableRes val iconUnselected: Int, @DrawableRes val iconSelected: Int, val title: String) {
     object BrainzPlayer : AppNavigationItem("brainzplayer", R.drawable.player_unselected, R.drawable.player_selected, "Player")
     object Explore : AppNavigationItem("explore", R.drawable.explore_unselected, R.drawable.explore_selected, "Explore")
-    object Profile : AppNavigationItem("profile", R.drawable.profile_unselected, R.drawable.profile_selected, "Profile")
+    object Profile : AppNavigationItem("profile", R.drawable.profile_unselected, R.drawable.profile_selected, "Profile") {
+        const val ARG_USERNAME = "username"
+
+        fun withUserArg(username: String): String {
+            return "$route?$ARG_USERNAME=$username"
+        }
+    }
     object Feed : AppNavigationItem("feed", R.drawable.feed_unselected, R.drawable.feed_selected, "Feed")
     object Settings: AppNavigationItem("settings", R.drawable.ic_settings, R.drawable.ic_settings, "Settings")
     object About: AppNavigationItem("about", R.drawable.ic_info, R.drawable.ic_info, "About")
--- a/app/src/main/java/org/listenbrainz/android/ui/navigation/AdaptiveNavigationBar.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/navigation/AdaptiveNavigationBar.kt
@@ -167,8 +167,7 @@ fun AdaptiveNavigationBar(
 
                     when (item.route) {
                         AppNavigationItem.Profile.route -> {
-                            val profileRoute = AppNavigationItem.Profile.route +
-                                    if (!username.isNullOrBlank()) "/${username}" else ""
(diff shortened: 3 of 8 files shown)


app/src/main/java/org/listenbrainz/android/ui/navigation/NavBarReorderOverlay.kt
```diff
@@ -0,0 +1,159 @@
...
+                    .border(3.dp, ListenBrainzTheme.colorScheme.text, CircleShape)
+                    .clickable {
+                        onDismiss(mutableItems.toList())
+                    },
+                contentAlignment = Alignment.Center
+            ) {
+                Icon(
+                    imageVector = Icons.Default.Close,
+                    contentDescription = "Close",
+                    tint = ListenBrainzTheme.colorScheme.text,
```
07jasjeet: ListenBrainzTheme.colorScheme.onLbSignature

app/src/main/java/org/listenbrainz/android/ui/navigation/ReorderableNavigationItems.kt
```diff
@@ -0,0 +1,123 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import org.listenbrainz.android.R
+import org.listenbrainz.android.ui.navigation.BottomNavItem
+import sh.calvin.reorderable.ReorderableItem
+import sh.calvin.reorderable.ReorderableLazyListState
+
+
+val navIcon: @Composable (BottomNavItem) -> Unit = { item ->
```
07jasjeet: This is just a function only written as a static getter, please convert it to function itself.

app/src/main/java/org/listenbrainz/android/ui/screens/main/MainActivity.kt
```diff
@@ -203,7 +203,8 @@ class MainActivity : ComponentActivity() {
                                     topBarActions = TopBarActions(),
                                     navigateToCreateAccount = {
                                         backStack.add(NavigationItem.CreateAccountScreen)
-                                    }
+                                    },
+                                    onReorderNav = {}
```
07jasjeet: This callback may be useful in future if it also provides the updated ordered list of nav items

app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsCallbacks.kt
```diff
@@ -25,5 +26,6 @@ data class SettingsCallbacksToHomeScreen(
     val topBarActions: TopBarActions,
     val onLoginRequest: () -> Unit,
     val onOnboardingRequest: () -> Unit,
-    val navigateToCreateAccount : () -> Unit
+    val navigateToCreateAccount : () -> Unit,
+    val onReorderNav: () -> Unit
```
07jasjeet: Should be named `onNavigationReorderClick` because the current name is slightly confusing and presents two possible behaviours:
1. Is this lambda being called AFTER the reorder? making it post reorder type callback.
2. The current behaviour.

rajeev-k47: Got it

app/src/main/java/org/listenbrainz/android/ui/screens/main/HomeScreen.kt
```diff
@@ -196,46 +209,51 @@ fun HomeScreen(
...
-                        topBarActions = topBarActions
+                        },
+                        snackbarState = snackbarState,
+                        dashBoardViewModel = dashBoardViewModel,
+                        topAppBarActions = topBarActions,
+                        settingsCallbacks = settingsCallbacks.copy(
+                            topBarActions = topBarActions,
+                            onReorderNav = {
+                                showNavReorderOverlay = true
+                            }
```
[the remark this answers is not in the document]
rajeev-k47: Fixed, I mistakenly used SettingsCallbacksToHomeScreen while it is not required

rajeev-k47:
I have done all the changes requested except the colorscheme fix.

07jasjeet (approved):
Hey, great work! Code looks good and clean. Most of the things have turned out exactly how we've wanted to. 

There's just one thing I've taken care of myself since the bug pre-dates this feature: When `Profile` tab is selected as first tab and we navigate back from overlay screen, the state of the profile tab is stuck in logged out state. This also happens on app launch.

You can refer to https://github.com/metabrainz/listenbrainz-android/pull/645/commits/0fdcb04b31fffb0c2238ebc3ed39dcc560ef0c56 on how it was resolved. To put it in short, we had a duplicated nav destination for profile route which only showed the login screen even though our normal profile screen handled both, logged in and logged out states. I have removed the dupe destination to fix the bug.

PR #679, opened 2026-01-18
Feature: Extending Search Functionality (Playlist, Artist, Album, Track)

### LINKED ISSUE #646: Add Playlist, Album, Artist, and Track search under the search bar
Currently, the ListenBrainz app supports only User Search through the search bar. To make search more flexible and consistent with the ListenBrainz Website, we can extend this functionality to support searching for playlists, albums, artists and tracks as well in the app, the corresponding endpoints are also available.

#### Existing search bar UX on the ListenBrainz website :- 

Image

nirvan73:
Fixes #646 

This PR extends the search functionality under the search bar in the LB app. It adds the respective network calls to the **ListenBrainz** and **MusicBrainz** API endpoints for playlist, album, artist, and track search. The query logic for each search type is implemented within the same `SearchViewModel`. Additionally, shimmer effect has been implemented for all search types during the loading phase, along with a minor card design update for the user search type. Clicking on a search result navigates the user to the corresponding playlist, album, artist, or user, for tracks the existing functionality of redirecting the user to Spotify or YouTube Music for the selected song has been retained. 

Screen Recording: -

https://github.com/user-attachments/assets/abfb7b0d-1e44-4baa-9924-1edc5241574c

Additional Note:- 
For playlist search, I added a custom `HttpClient` with a 30 second request and socket timeout. This is because the `LB_API/playlist/search` endpoint has higher latency, and without this timeout the request was failing early.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/navigation/AdaptiveNavigationBar.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/navigation/AdaptiveNavigationBar.kt
@@ -63,7 +63,6 @@ fun AdaptiveNavigationBar(
     currentlyPlayingSong: Song,
     listeningNowUIState: ListeningNowUIState,
     songList: List<Song>,
-    searchBarState: SearchBarState,
 ) {
     val coroutineScope = rememberCoroutineScope()
 
@@ -161,9 +160,6 @@ fun AdaptiveNavigationBar(
                         navController.popBackStack()
                     }
 
-                    if (searchBarState.isActive) {
-                        searchBarState.deactivate()
-                    }
 
                     when (item.route) {
                         AppNavigationItem.Profile.route -> {
@@ -249,7 +245,6 @@ fun AdaptiveNavigationBarPreview() {
         currentlyPlayingSong = Song(),
         listeningNowUIState = ListeningNowUIState(),
         songList = emptyList(),
-        searchBarState = rememberSearchBarState(),
         items = BottomNavDefaults.items()
     )
 }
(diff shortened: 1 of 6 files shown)


07jasjeet:
Hey, the implementation looks good but there are some changes that need to be done.

1. Currently the search screen is an overlay, it would be better if it was part of the navigation system as whenever I search for an entity (example an artist) and then hitting back, it does not revert to the search screen which is what user would expect. We need to make that possible.
2. We should use the same component for the "User", "Playlist" etc., tabs as the one being used in feed and profile.
3. When switching between tabs with the same query, we should not trigger the API again if the query stays the same. We can invalidate the cache if the query changes.
4. I think this is an existing bug, but when the user taps on or comes back and there is some query pre-filled, we should set the cursor position to the end of the text.

07jasjeet:
@nirvan73 For the second point, I do not mean the content cards but the selection tabs on the top below the search bar and above content.

nirvan73:
Ohh, got it. I’ll replace the selection tabs with the common selection tabs component used in the Feed and Profile screens.

Also just to clarify, the suggestion I mentioned earlier about creating a separate generic card component for User, Playlist, and Artist which was on my misunderstanding of the requirement, but I still wanted to check if you’d like me to proceed with that improvement as well or should i leave it as it is?

07jasjeet:
> Hey @07jasjeet, I came across an idea for the user search results, we can keep the top/first search result styled like these and rest the same?? Let me know your thoughts on whether we should do this, and if yes, which one to prefer.
> 
> image image

We should keep all items similar (Existing design). The UX tries to convey best match for query but the order of the list already conveys that. The way I see it, we are taking more screen space to convey the same thing.

PR #691, opened 2026-02-07
Handle THANKS feed event type to prevent feed failure

AyusKumarPathak:
Hi @07jasjeet 👋
I’ve raised a PR that adds support for resolving the new thanks feed event type so the Android feed no longer falls back to the “update app” error when such events appear.

Please let me know if this approach looks correct happy to follow up with a dedicated Thanks UI and original event rendering next.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/model/Metadata.kt
+++ b/app/src/main/java/org/listenbrainz/android/model/Metadata.kt
@@ -49,7 +49,27 @@ data class Metadata(
     val user0: String? = null,
     /** Used for follow following taglines. This is the one who was followed. */
     @SerialName("user_name_1")
-    val user1: String? = null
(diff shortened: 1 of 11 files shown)


07jasjeet (changes requested):
UI and API response model hookup is pending. Please complete the whole feature before requesting for review. Approving every commit or minor change will take unnecessary time, try to work async and get out a prototype at least before requesting review.

07jasjeet:
Hey @AyusKumarPathak, can you declare your usage of AI in this whole PR? Seems like you are just prompting and the LLM is just implementing to its best knowledge without you contributing any code.

Can you provide:
1. Estimated percentage of hand-written code.
2. Estimated percentage of hand-written documentation.
3. Commits that were committed by AI

PR #727, opened 2026-03-02
Implemented delete listen

### LINKED ISSUE #615: Add Delete Listens

anuj990:
This pr Fixes #615 
Implemented Delete Listen
Added DeleteListen data class,implemented ListensRepositoryImpl and ListensService ,
instant Listens remove,
Screen Recording

https://github.com/user-attachments/assets/d29ed6e2-af8f-420f-af06-085b0e300056


app/src/main/java/org/listenbrainz/android/service/ListensService.kt
```diff
@@ -43,5 +46,10 @@ interface ListensService {
     suspend fun getNowPlaying(
         @Path("user_name") username: String
     ): Listens
-    
+
+    @POST("delete-listen")
+    suspend fun deleteListen(
+        @Header("Authorization") token: String,
```
07jasjeet: We do not need to add auth header because this is automatically added to request.

app/src/main/java/org/listenbrainz/android/repository/listens/ListensRepositoryImpl.kt
```diff
@@ -77,8 +78,20 @@ class ListensRepositoryImpl(
...
     }
-    
-    
+    override suspend fun deleteListen(
+        token: String,
+        listenedAt: Long,
+        recordingMsid: String
+    ): Resource<PostResponse> = parseResponse {
+        service.deleteListen(
+            token = "Token $token",
```
07jasjeet: No need for token here.

app/src/main/java/org/listenbrainz/android/ui/screens/profile/listens/ListensScreen.kt
```diff
@@ -134,11 +127,13 @@ fun ListensScreen(
     snackbarState: SnackbarHostState,
     username: String?,
     goToArtistPage: (String) -> Unit,
-    goToUserProfile: (String) -> Unit
+    goToUserProfile: (String) -> Unit,
+    deleted: Map<Pair<Long, String>, Boolean>
```
07jasjeet: Lets use a data class instead of Pair class

app/src/main/java/org/listenbrainz/android/viewmodel/SocialViewModel.kt
```diff
@@ -238,9 +251,34 @@ class SocialViewModel(
...
+            val listenedAt = metadata.listenedAt
+            if(listenedAt == null){
+                return@launch
+            }
+            val msid = metadata.trackMetadata?.additionalInfo?.recordingMsid
+            if(msid ==null){
+                emitError(ResponseError.BadRequest())
+                return@launch
+            }
+            val result = listensRepository.deleteListen(token, listenedAt, msid)
```
07jasjeet: No token required here

07jasjeet (changes requested):
As far as implementation goes, we should not store which listens were deleted. We should keep it similar to desktop and show a pop up saying that the listen is scheduled for deletion. Refreshing should show the listen again. 

Also, I think we should follow up on progress of https://tickets.metabrainz.org/browse/LB-1141 and see if `staged_for_deletion` key is sent in Listen metadata.

app/src/main/java/org/listenbrainz/android/ui/screens/profile/listens/ListensScreen.kt
```diff
@@ -1424,7 +1448,8 @@ fun ListensScreenPreview() {
             getListensData = {},
             onFollowButtonClick = { _, _ -> },
             goToArtistPage = {},
-            goToUserProfile = {}
+            goToUserProfile = {},
+            deleted = {} as Map<Pair<Long, String>, Boolean>
```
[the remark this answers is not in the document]
anuj990: I will push the changes soon , 
And for the declaration of AI
I have not used AI in any of my PRs or commits so far, and I will not use it in the future too

app/src/main/java/org/listenbrainz/android/ui/screens/playlist/PlaylistDetailScreen.kt
```diff
@@ -535,10 +535,7 @@ private fun PlaylistDetailContent(
                         showsnackbar(error.toast)
                     },
                     goToArtistPage = goToArtistPage,
-                    onClick = {
-                        onTrackClick(playlistTrack)
```
07jasjeet: unnecessary/un-related change. Always preferred to have named arguments (over trailing lambda as well)

app/src/main/java/org/listenbrainz/android/ui/screens/profile/createdforyou/CreatedForYouScreen.kt
```diff
@@ -371,9 +371,6 @@ private fun CreatedForYouScreen(
                                         snackbarState.showSnackbar(error.toast)
                                     },
                                     goToArtistPage = goToArtistPage,
-                                    onClick = {
-                                        onTrackClick(playlist)
```
07jasjeet: Same here, unnecessary/un-related change. Its always preferred to have named arguments.

07jasjeet:
Thanks for the contribution @anuj990!

Issue #660, opened 2026-01-11
Replace Logger-Android with Kermit or Napier


07jasjeet:
Switch logging from Logger-Android to a multiplatform logging library like Kermit or Napier.

javaman97:
**Kermit** and **Napier**, two Kotlin Multiplatform–compatible logging libraries, from a **feature** and **maintenance** perspective.  

## Feature-wise Comparison

### Kotlin Multiplatform Support

- **Kermit**
  - Designed Kotlin Multiplatform–first
  - Supports Android, iOS, JVM, JS, and Native
  - Consistent API and behavior across platforms

- **Napier**
  - Supports Kotlin Multiplatform (Android, iOS, JVM, JS)
  - Platform handling is implemented via Antilog abstractions

---

### Logging Capabilities

- **Kermit**
  - Supports full log severity levels
  - Enables structured logging
  - Supports tagging and metadata
  - Lazy message evaluation for improved performance

- **Napier**
  - Supports basic log levels
  - No built-in structured logging
  - Messages are evaluated eagerly

---

### Extensibility

- **Kermit**
  - Pluggable log writers (Logcat, NSLog, Console, custom)
  - Easy to extend for custom outputs or formats
  - Writers can be replaced or mocked for testing

- **Napier**
  - Uses Antilog for output customization
  - Extensibility is more limited compared to Kermit
  - Testing requires additional setup

---

### API Design & Usability

- **Kermit**
  - Clean, expressive API
  - Similar to Android logging patterns
  - Suitable for both simple and advanced use cases

- **Napier**
  - Very simple and lightweight API
  - Well-suited for minimal logging needs

---

## Maintenance-wise Comparison

### Maintenance Activity

- **Kermit**
  - Actively maintained by Touchlab
  - Regular releases and updates
  - Ongoing feature enhancements

- **Napier**
  - Maintained by a smaller contributor group
  - Slower release cadence
  - Fewer recent updates

---

### Community & Ecosystem

- **Kermit**
  - Widely adopted in Kotlin Multiplatform projects
  - Strong community presence
  - Frequently recommended in KMP ecosystem

- **Napier**
  - Moderate adoption
  - Smaller ecosystem compared to Kermit

---

### Documentation & Support

- **Kermit**
  - Comprehensive documentation
  - Clear examples and usage guides

- **Napier**
  - Basic documentation
  - Limited advanced usage examples

---

### Long-term Viability

- **Kermit**
[...]

PR #728, opened 2026-03-02
Replace Logger-Android with Kermit

javaman97:
fixes : #660 

What I implemented till now ->
- Added Kermit dependency in shared module
- Designed shared Log 
- import shared Log in app module

- Remove Log.kt of Android
- Remove Android Logger Dependency in build.gradle(module)
- platform- specific for logger 

_Log Sharing ->  compressToZip -> send via email_
` 

> fun submitLogs(context: Context) {
>         Logger.apply {
>             compressLogsInZipFile { zipFile ->
>                 zipFile?.let {
>                     FileIntent
>                         .fromFile(
>                             context,
>                             it,
>                             BuildConfig.APPLICATION_ID
>                         )
>                         ?.let { intent ->
>                             intent.putExtra(Intent.EXTRA_SUBJECT, "Log Files")
>                             intent.putExtra(
>                                 Intent.EXTRA_EMAIL,
>                                 arrayOf("mobile@metabrainz.org")
>                             )
>                             intent.putExtra(
>                                 Intent.EXTRA_TEXT,
>                                 "Please find the attached log files."
>                             )
>                             intent.putExtra(
>                                 Intent.EXTRA_STREAM,
>                                 FileProvider.getUriForFile(
>                                     context,
>                                     "${BuildConfig.APPLICATION_ID}.provider",
>                                     zipFile
>                                 )
>                             )
>                             intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
>                             try {
>                                 context.startActivity(
>                                     Intent.createChooser(
>                                         intent,
>                                         "Email logs..."
>                                     )
>                                 )
>                             } catch (e: java.lang.Exception) {
>                                 e(throwable = e)
>                             }
>                         }
>                 }
>             }
>         }
>     }

`
(as it provided by Android logger, @07jasjeet should I implement zip file logic manually ? please share your valuable feedback )

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/ConsentWebViewClient.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/ConsentWebViewClient.kt
@@ -3,7 +3,7 @@ package org.listenbrainz.android.ui.screens.onboarding.auth
 import android.graphics.Bitmap
 import android.webkit.WebView
 import android.webkit.WebViewClient
-import org.listenbrainz.shared.util.Log
+import org.listenbrainz.android.util.Log
 
 class ConsentWebViewClient(
     private val onLoadData: (String) -> Unit,
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/ConsentWebViewClient.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/ConsentWebViewClient.kt
@@ -3,7 +3,7 @@ package org.listenbrainz.android.ui.screens.onboarding.auth
 import android.graphics.Bitmap
 import android.webkit.WebView
 import android.webkit.WebViewClient
-import com.limurse.logger.Logger
+import org.listenbrainz.shared.util.Log
 
 class ConsentWebViewClient(
     private val onLoadData: (String) -> Unit,
@@ -15,14 +15,14 @@ class ConsentWebViewClient(
     override fun onPageStarted(view: WebView?, url: String?, favicon: Bitmap?) {
         super.onPageStarted(view, url, favicon)
         hadError = false
-        Logger.d(TAG, "Page started loading: $url")
+        Log.d("Page started loading: $url", tag = TAG)
     }
 
 
     override fun onPageFinished(view: WebView?, url: String?) {
         super.onPageFinished(view, url)
         if (hadError) return
-        Logger.d(TAG, "Page finished loading: $url")
+        Log.d("Page finished loading: $url", tag = TAG)
 
         if (url?.contains("listenbrainz.org/login") == true) {
             view?.postDelayed(
@@ -45,11 +45,11 @@ class ConsentWebViewClient(
                     
                 """.trimIndent()
                     ) { value ->
-                        Logger.d(TAG, "Consent Screen Data: $value")
+                        Log.d("Consent Screen Data: $value", tag = TAG)
                         if (value != null && value != "null" && !value.startsWith("Error:")) {
                             onLoadData(value)
                         } else {
-                            Logger.d(
+                            Log.d(
                                 TAG,
                                 "Failed to retrieve consent screen data or no consent text found."
                             )
--- a/app/src/main/java/org/listenbrainz/android/util/Log.kt
+++ b/app/src/main/java/org/listenbrainz/android/util/Log.kt
@@ -1,25 +1,25 @@
 package org.listenbrainz.android.util
 
-import com.limurse.logger.Logger
+import org.listenbrainz.shared.util.Log
 
-interface Log {
-    fun e(message: Any?, tag: String? = null)
-    
-    fun d(message: Any?)
-    
-    fun w(message: Any?)
-    
-    companion object: Log {
-        override fun e(message: Any?, tag: String?) {
-            Logger.e(tag, msg = message.toString())
-        }
-        
-        override fun d(message: Any?) {
-            Logger.d(msg = message.toString())
-        }
-        
-        override fun w(message: Any?) {
-            Logger.w(msg = message.toString())
-        }
-    }
-}
+//interface Log {
+//    fun e(message: Any?, tag: String? = null)
+//
+//    fun d(message: Any?)
+//
+//    fun w(message: Any?)
+//
+//    companion object: Log {
+//        override fun e(message: Any?, tag: String?) {
+//            Logger.e(tag, msg = message.toString())
+//        }
+//
+//        override fun d(message: Any?) {
+//            Logger.d(msg = message.toString())
+//        }
+//
+//        override fun w(message: Any?) {
+//            Logger.w(msg = message.toString())
+//        }
+//    }
+//}
--- a/app/src/main/java/org/listenbrainz/android/util/Log.kt
+++ b/app/src/main/java/org/listenbrainz/android/util/Log.kt
@@ -1,25 +1,138 @@
 package org.listenbrainz.android.util
 
+import android.content.Context
+import android.os.Build
+import org.listenbrainz.android.BuildConfig
 import org.listenbrainz.shared.util.Log
+import java.io.File
+import java.io.FileWriter
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale.getDefault
 
-//interface Log {
-//    fun e(message: Any?, tag: String? = null)
-//
-//    fun d(message: Any?)
-//
-//    fun w(message: Any?)
-//
-//    companion object: Log {
-//        override fun e(message: Any?, tag: String?) {
-//            Logger.e(tag, msg = message.toString())
-//        }
-//
-//        override fun d(message: Any?) {
-//            Logger.d(msg = message.toString())
-//        }
-//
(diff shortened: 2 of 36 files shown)


app/src/main/java/org/listenbrainz/android/util/Log.kt
```diff
@@ -1,25 +1,138 @@
...
+                writer.write(logEntry)
+                writer.flush()
+            }
+        } catch (e: Exception) {
+            Log.e("LogFileWriter", "Failed to write to file", e)
+        }
+    }
+
+    fun e(message: Any?, tag: String? = null, throwable: Throwable? = null) {
+        Log.e(message, tag, throwable)
```
07jasjeet, line 85: **Critical: Infinite recursion / StackOverflowError**

`Log.e(message, tag, throwable)` here resolves to `org.listenbrainz.android.util.Log.e` — i.e., this very method — because `object Log` in this file shadows the `import org.listenbrainz.shared.util.Log` at line 6. Every call to `Log.e/d/w/i/v` inside this object is a recursive call to itself.

Fix: use a named import alias or qualify the call explicitly:
```kotlin
import org.listenbrainz.shared.util.Log as SharedLog

fun e(message: Any?, tag: String? = null, throwable: Throwable? = null) {
    SharedLog.e(message, tag, throwable)
    ...
}
```
The same issue applies to every delegating call in `d`, `w`, `i`, `v`, and the `catch` blocks in `init` and `collectStartupData`.

app/src/main/java/org/listenbrainz/android/util/Log.kt
```diff
@@ -1,25 +1,138 @@
...
+
+    private fun writeToFile(
+        message: String,
+        tag: String = "ListenBrainz",
+        severity: String = "INFO",
+    ) {
+        if (!isInitialized) return
+
+        try {
+            FileWriter(logFile, true).use { writer ->
```
07jasjeet, line 73: **Bug: Synchronous file I/O on the calling thread**

`FileWriter` is a blocking call. `writeToFile` is invoked from `Log.d/e/w/i/v`, which are called from ViewModels, WebViewClients, and Composables — all on the main thread. This will trigger StrictMode violations and can cause ANRs.

Consider either:
- Using a `BufferedWriter` backed by a background-thread dispatcher (e.g. a single-threaded `Dispatchers.IO` executor), or
- Batching writes via a Channel/Queue on a dedicated IO coroutine.

07jasjeet, line 73: **Bug: No synchronization on concurrent file writes**

Multiple threads can call `writeToFile` concurrently. Each call opens a new `FileWriter(logFile, true)` independently — there is no mutex or synchronization, so log lines from different threads will interleave and potentially corrupt the file. The old library handled this internally.

A `Mutex` around the write, or routing all writes through a single-threaded coroutine dispatcher, would fix this.

app/src/main/java/org/listenbrainz/android/util/Log.kt
```diff
@@ -1,25 +1,138 @@
...
+    }
+
+    fun v(message: Any?, tag: String? = null) {
+        Log.v(message, tag)
+
+        val logTag = tag ?: "ListenBrainz"
+        writeToFile(message.toString(), logTag, "VERBOSE")
+    }
+
+    fun log(message: String, tag: String = Constants.TAG, severity: String = "INFO") {
```
07jasjeet, line 122: **Bug: `log()` bypasses logcat entirely**

This public method writes only to file and never calls through to Kermit/logcat, unlike `e/d/w/i/v` which (once the recursion bug is fixed) will delegate to `SharedLog`. Any caller using `Log.log()` will get silent file-only logging with no logcat output.

Either remove it (it duplicates `writeToFile`) or route it through the shared logger the same way as the other methods.

07jasjeet (changes requested):
Overall direction is good — Kermit is a solid KMP-friendly logger. However there are several critical bugs that will cause crashes at runtime. See inline comments.

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/ConsentWebViewClient.kt
```diff
@@ -45,11 +45,11 @@ class ConsentWebViewClient(
                     
                 """.trimIndent()
                     ) { value ->
-                        Logger.d(TAG, "Consent Screen Data: $value")
+                        Log.d("Consent Screen Data: $value", tag = TAG)
                         if (value != null && value != "null" && !value.startsWith("Error:")) {
                             onLoadData(value)
                         } else {
-                            Logger.d(
+                            Log.d(
```
07jasjeet, line 52: **Bug: Arguments are swapped — TAG logged as the message, message logged as the tag**

The new `Log.d` signature is `d(message: Any?, tag: String? = null)`. Here `TAG` is passed as the first positional argument (message) and the actual log text as the second (tag), which is the reverse of intent.

Compare to the correctly named calls just above in the same file:
```kotlin
Log.d("Consent Screen Data: $value", tag = TAG)  // correct
```
Fix:
```kotlin
Log.d("Failed to retrieve consent screen data or no consent text found.", tag = TAG)
```

app/src/main/java/org/listenbrainz/android/util/Utils.kt
```diff
@@ -680,48 +682,45 @@ object Utils {
...
+                logDir.listFiles()?.forEach { file ->
+                    if (file.isFile && file.extension == "txt") {
+                        val entry = ZipEntry(file.name)
+                        zos.putNextEntry(entry)
+                        file.inputStream().use { it.copyTo(zos) }
+                        zos.closeEntry()
+                    }
+                }
+            }
+
```
07jasjeet, line 702: **Critical: `startActivity` called on an IO thread**

`context.startActivity` is a UI operation that must run on the main thread. This `withContext(Dispatchers.IO)` block is nested inside an outer `withContext(Dispatchers.IO)` (the function signature), so `startActivity` a few lines below will execute on an IO thread and throw a `CalledFromWrongThreadException` at runtime.

Fix: switch to `Dispatchers.Main` for the intent dispatch:
```kotlin
// zip creation stays on IO
ZipOutputStream(...).use { ... }

// then switch to Main to launch the chooser
withContext(Dispatchers.Main) {
    context.startActivity(Intent.createChooser(intent, "Email logs..."))
}
```

shared/src/androidMain/kotlin/org/listenbrainz/shared/Platform.android.kt
```diff
@@ -1,3 +1,24 @@
...
+
+actual fun provideLogger(): Logger {
+
+    val writers = mutableListOf(
+        platformLogWriter()
+    )
+
+    return Logger(
+        config = StaticConfig(
+            minSeverity = Severity.Info,
```
07jasjeet, line 19: **Bug: `Severity.Info` silently drops all Debug and Verbose logs**

With `minSeverity = Severity.Info`, every `Log.d()` and `Log.v()` call in the codebase (and there are many, including the ones migrated in this PR) will produce no logcat output. This makes debugging very difficult.

Consider gating on build type:
```kotlin
minSeverity = if (BuildConfig.DEBUG) Severity.Debug else Severity.Info
```

shared/src/androidMain/kotlin/org/listenbrainz/shared/Platform.android.kt
```diff
@@ -1,3 +1,24 @@
...
\ No newline at end of file
+import co.touchlab.kermit.Logger
+import co.touchlab.kermit.Severity
+import co.touchlab.kermit.StaticConfig
+import co.touchlab.kermit.platformLogWriter
+
+actual fun platform() = "Android"
+
+
+actual fun provideLogger(): Logger {
```
07jasjeet, line 11: **`provideLogger()` is declared but never called anywhere**

This `expect/actual` function is defined for both Android and iOS platforms but the shared `Log` companion (in `shared/src/commonMain/kotlin/org/listenbrainz/shared/util/Log.kt`) creates its own `Logger` instance directly instead of calling `provideLogger()`. As a result this whole expect/actual mechanism is dead code.

Either wire `provideLogger()` into the shared `Log` companion, or remove the expect/actual declaration and the platform implementations.

app/src/main/java/org/listenbrainz/android/util/Log.kt
```diff
@@ -1,25 +1,138 @@
...
+            Log.e("LogFileWriter", "Failed to initialize file logging", e)
         }
-        
-        override fun d(message: Any?) {
-            Logger.d(msg = message.toString())
+    }
+
+    private fun collectStartupData() {
+        val startupData = mapOf(
+            "App Version" to System.currentTimeMillis().toString(),
```
07jasjeet, line 41: **Pre-existing bug: `"App Version"` maps to a timestamp, not the version**

`System.currentTimeMillis().toString()` produces a Unix timestamp in milliseconds, not the app version string. This was carried over from the old `collectStartupData()` in `App.kt`. Since this function is being rewritten here, worth fixing:
```kotlin
"App Version" to BuildConfig.VERSION_NAME
```

07jasjeet:
Closing as completed by: https://github.com/metabrainz/listenbrainz-android/pull/742

PR #731, opened 2026-03-07
Migrate socket.io-client to ktor-client-websockets in shared module

### LINKED ISSUE #666: Replace socket.io-client with ktor-client-websockets
Switch realtime/data streaming from socket.io-client to ktor-client-websockets, ensuring multiplatform support.

nirvan73:
This PR fixes #666 by migrating the realtime communication layer from `socket.io-client` to `ktor-client-websockets` to support KMP. Since `socket.io-client` lacks KMP compatibility, this impl manually handles the Engine.io/Socket.io protocol within `SocketRepositoryImpl`. I have implemented the manual handshake by sending "40" code to init the session, the heartbeat mechanism by responding with "3" to incoming "2" code to keep the session alive, and event subscription for JSON payloads. The change allows the realtime events "lIsten" and "playing_now" to be shared across the platforms i.e Android and iOS in the `commonMain` package under shared module.

Additional:-
Migrated `Listen` and related data classes in the shared module to avoid duplication and unify the data layer for both platforms.


app/build.gradle.kts
```diff
@@ -137,6 +137,7 @@ android {
 
     lint {
         abortOnError = false
+        ignoreTestSources = true
```
07jasjeet, line 140: Reason for this?

shared/src/commonMain/kotlin/org/listenbrainz/shared/repository/socket/SocketRepositoryImpl.kt
```diff
@@ -0,0 +1,64 @@
...
+) : SocketRepository {
+
+    private val json = Json { ignoreUnknownKeys = true }
+
+    override fun listen(usernameProvider: suspend () -> String) = callbackFlow {
+        val username = usernameProvider()
+        httpClient.webSocket(
+            urlString = "wss://listenbrainz.org/socket.io/?EIO=4&transport=websocket"
+        ) {
+            send(Frame.Text("40"))
```
07jasjeet: I would rather use some kind of KMP library that implements Socket IO for simplicity. Maybe try https://github.com/HackWebRTC/kmp-socketio or implement minimal use-case related (listening to events) socket.io implementation that can then be used elsewhere.

.github/workflows/releaseProd.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'oracle'
-          java-version: 17
+          java-version: 21
```
07jasjeet, line 22: This is already done on prod, take pull.

.github/workflows/releaseBeta.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'oracle'
-          java-version: 17
+          java-version: 21
```
07jasjeet, line 20: https://github.com/metabrainz/listenbrainz-android/pull/731/changes#r3179557320

.github/workflows/attach-release.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'oracle'
-          java-version: 17
+          java-version: 21
```
07jasjeet, line 19: https://github.com/metabrainz/listenbrainz-android/pull/731/changes#r3179557320

.github/workflows/android.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'oracle'
-          java-version: 17
+          java-version: 21
```
07jasjeet, line 20: https://github.com/metabrainz/listenbrainz-android/pull/731/changes#r3179557320

07jasjeet (changes requested):
Need to take pull of main here and resolve merge conflicts. Some comments to address, mostly looks good.

app/build.gradle.kts
nirvan73: i believe this change came in from the cmp branch merge https://github.com/metabrainz/listenbrainz-android/commit/1414ac0fc373c1b11740279b5d1c30c6470bb1ea when i pulled the updated branch ?

shared/src/commonMain/kotlin/org/listenbrainz/shared/repository/socket/SocketRepositoryImpl.kt
nirvan73: i have switched to [kmp-socketio](https://github.com/HackWebRTC/kmp-socketio) as suggested, as it handles the Socket IO protocol like handshake and etc itself across all the KMP targets.

07jasjeet (commented):
## Review: Migrate socket.io-client to ktor-client-websockets

The goal is valid (KMP compatibility), but there are several issues worth addressing.

---

### Bugs / Correctness

**1. Unused imports in `SocketRepositoryImpl.kt`**

```kotlin
import io.ktor.client.plugins.websocket.webSocket  // never called
import io.ktor.websocket.Frame                      // never called
import io.ktor.websocket.readText                   // never called
```

`kmp-socketio` uses the Ktor `HttpClient` internally as transport, but these Ktor WebSocket APIs are never called directly. Dead imports should be removed.

---

**2. Listener leak on `awaitClose`**

Old code:
```kotlin
awaitClose {
    socket.disconnect()
    socket.off()  // removes ALL listeners
}
```

New code:
```kotlin
awaitClose {
    activeSocket?.close()  // does this also call off()?
}
```

If `kmp-socketio`'s `close()` doesn't remove registered listeners and the flow is collected again (e.g. after a reconnect), listeners will accumulate. Needs verification or an explicit `socket.off()` call.

---

**3. `activeSocket` null risk in `awaitClose`**

```kotlin
var activeSocket: Socket? = null

IO.socket("https://listenbrainz.org", options) { socket ->
    activeSocket = socket
    socket.open()
}

awaitClose {
    activeSocket?.close()  // null if callback never ran
}
```

If the `IO.socket` callback throws before `activeSocket` is assigned, cleanup is silently skipped.

---

**4. `Listen.kt` — removed `Metadata`/`TrackMetadata` imports**

```diff
-import org.listenbrainz.shared.model.Metadata
-import org.listenbrainz.shared.model.TrackMetadata
```

If `Listen` still has fields of those types, this won't compile. The diff doesn't show the full class body — needs confirmation that those fields were intentionally removed.

---

### Code Quality

**5. `val httpClient` should be `private val`**

```kotlin
class SocketRepositoryImpl(
    val httpClient: HttpClient  // should be private
) : SocketRepository {
```

It's an internal implementation detail, not part of the interface contract.

---

**6. `println` for error logging**

```kotlin
println("SocketRepository: $event error ${e.message}")
```

The old code used the project's `Log` utility. Bare `println` is not appropriate for production — consider injecting a logger or using a KMP-compatible logging abstraction.

---

**7. Self-shadowing variable in `ListeningNowViewModel.kt`**

```kotlin
val listenedAt = listenedAt  // shadows the class property
```

This works for null-smart-casting but is confusing. Use a distinct name like `val capturedListenedAt = listenedAt`.

---

**8. WebSockets plugin installed on the shared API `HttpClient`**

```kotlin
install(WebSockets)  // added to the base client used for all REST API calls
```

This works, but adds WebSocket overhead to a client otherwise used only for HTTP REST. A separate dedicated `HttpClient` for the socket connection would be cleaner.

---

### Inaccurate PR Description

**9. The description doesn't match the implementation**

> "manually handles the Engine.io/Socket.io protocol. Key implementations include: Manual handshake sending '40' code... Heartbeat mechanism responding with '3' to incoming '2' code"

None of this is in the code — the implementation delegates entirely to `com.piasy.kmp.socketio`, which handles the protocol internally. The description reads like a prior iteration and should be updated to avoid misleading reviewers.

---

### Library Choice

**10. `com.piasy.kmp.socketio` is a community library**

This is not an official socket.io KMP client — it's a lesser-known library by a single maintainer. Worth a brief discussion on whether it's production-ready and actively maintained before accepting it as a dependency.

---

### Minor notes

- The `offsetOrZero()` fix in `BrainzPlayerBackDropScreen` is reasonable. Silent exception swallow is slightly concerning (no log on failure), but functionally acceptable.
- Moving `CoverArt`, `Image`, `Thumbnails` to the shared module is clean.
- Version bump (`patch 0 → 1`) may conflict — `main` already has `Prepare v2.15.1.0` as the latest commit.

07jasjeet (commented):
## Review: Migrate socket.io-client to ktor-client-websockets

The goal is valid (KMP compatibility), but there are several issues worth addressing.

---

### Bugs / Correctness

**1. Unused imports in `SocketRepositoryImpl.kt`**

```kotlin
import io.ktor.client.plugins.websocket.webSocket  // never called
import io.ktor.websocket.Frame                      // never called
import io.ktor.websocket.readText                   // never called
```

`kmp-socketio` uses the Ktor `HttpClient` internally as transport, but these Ktor WebSocket APIs are never called directly. Dead imports should be removed.

---

**2. Listener leak on `awaitClose`**

Old code:
```kotlin
awaitClose {
    socket.disconnect()
    socket.off()  // removes ALL listeners
}
```

New code:
```kotlin
awaitClose {
    activeSocket?.close()  // does this also call off()?
}
```

If `kmp-socketio`'s `close()` doesn't remove registered listeners and the flow is collected again (e.g. after a reconnect), listeners will accumulate. Needs verification or an explicit `socket.off()` call.

---

**3. `activeSocket` null risk in `awaitClose`**

```kotlin
var activeSocket: Socket? = null

IO.socket("https://listenbrainz.org", options) { socket ->
    activeSocket = socket
    socket.open()
}

awaitClose {
    activeSocket?.close()  // null if callback never ran
}
```

If the `IO.socket` callback throws before `activeSocket` is assigned, cleanup is silently skipped.

---

**4. `Listen.kt` — removed `Metadata`/`TrackMetadata` imports**

```diff
-import org.listenbrainz.shared.model.Metadata
-import org.listenbrainz.shared.model.TrackMetadata
```

If `Listen` still has fields of those types, this won't compile. The diff doesn't show the full class body — needs confirmation that those fields were intentionally removed.

---

### Code Quality

**5. `val httpClient` should be `private val`**

```kotlin
class SocketRepositoryImpl(
    val httpClient: HttpClient  // should be private
) : SocketRepository {
```

It's an internal implementation detail, not part of the interface contract.

---

**6. `println` for error logging**

```kotlin
println("SocketRepository: $event error ${e.message}")
```

The old code used the project's `Log` utility. Bare `println` is not appropriate for production — consider injecting a logger or using a KMP-compatible logging abstraction.

---

**7. Self-shadowing variable in `ListeningNowViewModel.kt`**

```kotlin
val listenedAt = listenedAt  // shadows the class property
```

This works for null-smart-casting but is confusing. Use a distinct name like `val capturedListenedAt = listenedAt`.

---

**8. WebSockets plugin installed on the shared API `HttpClient`**

```kotlin
install(WebSockets)  // added to the base client used for all REST API calls
```

This works, but adds WebSocket overhead to a client otherwise used only for HTTP REST. A separate dedicated `HttpClient` for the socket connection would be cleaner.

---

### Inaccurate PR Description

**9. The description doesn't match the implementation**

> "manually handles the Engine.io/Socket.io protocol. Key implementations include: Manual handshake sending '40' code... Heartbeat mechanism responding with '3' to incoming '2' code"

None of this is in the code — the implementation delegates entirely to `com.piasy.kmp.socketio`, which handles the protocol internally. The description reads like a prior iteration and should be updated to avoid misleading reviewers.

---

### Library Choice

**10. `com.piasy.kmp.socketio` is a community library**

This is not an official socket.io KMP client — it's a lesser-known library by a single maintainer. Worth a brief discussion on whether it's production-ready and actively maintained before accepting it as a dependency.

---

### Minor notes

- The `offsetOrZero()` fix in `BrainzPlayerBackDropScreen` is reasonable. Silent exception swallow is slightly concerning (no log on failure), but functionally acceptable.
- Moving `CoverArt`, `Image`, `Thumbnails` to the shared module is clean.
- Version bump (`patch 0 → 1`) may conflict — `main` already has `Prepare v2.15.1.0` as the latest commit.

nirvan73:
Hey @07jasjeet, i have made the changes asked in review:-
1. removed unused imports 
5. made `httpClient` private
6. Regarding the `println`, it will be replaced in the future when we move to a kmp supported logging library such as Kermit.

shared/src/commonMain/kotlin/org/listenbrainz/shared/repository/socket/SocketRepositoryImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.put
+import org.listenbrainz.shared.model.Listen
+
+
+class SocketRepositoryImpl(
+    private val httpClient: HttpClient
+) : SocketRepository {
+
+    private val json = Json { ignoreUnknownKeys = true }
```
07jasjeet: Can we use dependency injected json instead of creating a new one here. If our DI does not have a json config provider, lets add that.

07jasjeet (changes requested):

shared/src/commonMain/kotlin/org/listenbrainz/shared/repository/socket/SocketRepositoryImpl.kt
nirvan73: added a json config provider under di config file and injected it in this file 👍

07jasjeet (approved):

PR #732, opened 2026-03-07
Migrated SocialViewModel to shared

### LINKED ISSUE #701: Migrate SocialViewModel and dependencies to KMP shared module
Migrate all logic and dependencies of the `SocialViewModel` into the KMP shared module. Adapt code to ensure platform-agnostic compatibility and updating dependencies.

anuj990:
this PR fixes #701 

Moved social domain models  SocialData, RecommendationData, Resource, ResponseError, etc. to the shared module

Added SocialStateHolder to replace android ViewModel logic with coroutine based state management


07jasjeet:
Hey, merge conflicts need to be resolved + if files are directly moved, they need to preserve their commit history. Any changes to any files should be done after committing their move to another folder/rename.

07jasjeet (changes requested):
While I appreciate the work, I would really like to retain commit history of files and move the related files rather than duplicating them. The file transfer should be registered by git such that the commit history is preserved. Is it possible to do that at this stage? @nirvan73 can help.

anuj990:
> While I appreciate the work, I would really like to retain commit history of files and move the related files rather than duplicating them. The file transfer should be registered by git such that the commit history is preserved. Is it possible to do that at this stage? @nirvan73 can help.

Sure, it would be good to get some help with this.

07jasjeet:
Hi, thank you for the time dedicated and the work, but we will have close this PR. We could not build upon your work and starting from scratch was easier.

PR #735, opened 2026-03-08
Fixes feed overlap when ErrorBar appears #734

### LINKED ISSUE #734: [Bug] Feed screen layout overlap when ErrorBar appears
**Bug**
When a network error happens on the Feed screen, an **ErrorBar** shows up above the feed. When this appears, the header section becomes taller (TopBar + ErrorBar + NavigationChips), but the feed list below does not shift properly. Because of this, the feed cards move up and overlap the header area.

**Steps to Reproduce** => Open the Feed screen first with network, then turn off network and pull refresh, when the ErrorBar appears, look at the first few feed items.

**Cause**
The feed list currently uses a fixed spacer (StartingSpacer) at the top to account for the header height. When the ErrorBar appears, the header becomes taller but the spacer stays the same height, which causes the overlap.

rajeev-k47:
Fixes #734 

https://github.com/user-attachments/assets/179db080-0095-47ab-a4a9-12e0898106c2

Fixes the layout overlap on the Feed screen when the ErrorBar appears. Previously, when a network error occurred, the ErrorBar increased the height of the header section (TopBar + ErrorBar + NavigationChips), but the feed list below did not adjust its starting offset. This caused feed tiles to overlap with the header area.

The fix makes the spacer at the top of the feed list dynamic instead of fixed. Its height now depends on whether the ErrorBar is visible. When the error banner appears, the spacer becomes larger so the feed content starts lower and remains correctly positioned below the header. When there is no error, the spacer keeps the normal height.

This keeps the layout consistent and prevents the overlap issue across all feed tabs (My Feed, Follow Listens, and Similar Listens).

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/screens/feed/FeedScreen.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/screens/feed/FeedScreen.kt
@@ -238,6 +238,7 @@ fun FeedScreen(
                         listState = myFeedListState,
                         pagingData = myFeedPagingData,
                         uiState = uiState.myFeedState,
+                        errorDialogVisible = uiState.error != null,
                         onDeleteOrHide = callbacks.onDeleteOrHide,
                         recommendTrack = callbacks.onRecommend,
                         personallyRecommendTrack = { index ->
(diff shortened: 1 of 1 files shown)


07jasjeet (changes requested):

rajeev-k47:
@07jasjeet I have done the changes requested, now instead of hard coded padding it dynamically adds the combined padding of ErrorBarHeight and NavigationChipsHeight to the content. Tested on different DPI settings.

07jasjeet (approved):
Thanks for the contribution @rajeev-k47

PR #736, opened 2026-03-10
Fix : multiple clicks on settings icon open song player

### LINKED ISSUE #722: [Bug] : Multiple or frequent clicks on Settings Icon opens the song Player screen
**Steps to Reproduce**

- Open the app.
- Navigate to the screen where the Settings icon is visible.
- Rapidly tap (multiple times) on the Settings icon.
- Observe the navigation behavior.

**Expected Result**

- Only the Settings screen should open.
- Multiple rapid clicks should not trigger unintended navigation.

https://github.com/user-attachments/assets/2602a82c-2104-4161-9e25-e550ef41fc16

javaman97:
fixes : #722 

I have fixed the issue by using debouncing.

### DIFF
--- a/app/src/main/java/org/listenbrainz/android/ui/navigation/TopBar.kt
+++ b/app/src/main/java/org/listenbrainz/android/ui/navigation/TopBar.kt
@@ -11,6 +11,12 @@ import androidx.compose.material.TopAppBar
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/listenbrainz/android/ui/navigation/TopBar.kt
```diff
@@ -36,6 +42,16 @@ fun TopBar(
     context: Context = LocalContext.current,
     topBarActions: TopBarActions
 ) {
+
+    var isSettingsClickEnabled by remember { mutableStateOf(true) }
+
+    LaunchedEffect(isSettingsClickEnabled) {
+        if (!isSettingsClickEnabled) {
+            delay(500)
```
07jasjeet, line 50: Adding random delay seems like a hacky solution, need to somehow observe navController and await transition of navigation.

07jasjeet:
Closing due to inactivity.

PR #742, opened 2026-05-28
Continue logger-kermit migration and address review feedback

nirvan73:
This PR fixes the reviews left in #728 and continue the migration of Logger to Kermit to support KMP.
###  Changes made for code reviews:
1. Resolved the recursive call inside the Android `Log` provider by using a `sharedLog` import alias.
2. Migrated disk write operations away from the main thread to a thread-safe background Channel on a dedicated IO coroutine.
3. Resolved the concurrent file write bug via this same `Channel`, it completely eliminates the need for a Mutex lock and prevents multiple threads from corrupting the file.
4. Resolved the `log()` bypass bug by completely removing the redundant public `log()` method from the Android utility class.
5. Fixed the swapped log arguments in `ConsentWebViewClient` by explicitly passing `tag = TAG`.
6. Wrapped `context.startActivity` in `withContext(Dispatchers.Main)` to safely run the UI intent on the main thread instead of the IO thread
7. Changed the log minSeverity to `Debug` to stop local logs from dropping silently under Platform Specific impl.
8. Replaced the manual init of the logger with `provideLogger` into the shared `Log` object so platform settings work


nirvan73:
Hey @07jasjeet, I have done the fixes for the reviews left in #728. Please have a look at them

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/introduction/IntroductionScreens.kt
```diff
@@ -334,11 +337,14 @@ fun OnboardingBackButton(modifier: Modifier = Modifier, onBackPress: (() -> Unit
...
     val haptic = LocalHapticFeedback.current
     val context = LocalContext.current
+    val scope = rememberCoroutineScope()
     IconButton(
         modifier = modifier,
         onClick = {
             haptic.performHapticFeedback(HapticFeedbackType.Confirm)
-            submitLogs(context)
+            scope.launch {
+                submitLogs(context)
```
07jasjeet: we need to handle this in a scope that survives configuration change, like `viewModelScope`. Also, we need to show some kind of loader for user information and block multiple touches such that submitLogs is not called repeatedly.

app/src/main/java/org/listenbrainz/android/ui/screens/settings/SettingsScreen.kt
```diff
@@ -314,7 +314,9 @@ fun SettingsScreen(
 
             SettingsTextOption(
                 modifier = Modifier.clickable {
-                    submitLogs(context)
+                    scope.launch {
+                        submitLogs(context)
```
07jasjeet: https://github.com/metabrainz/listenbrainz-android/pull/742/changes#r3329797303

app/src/main/java/org/listenbrainz/android/util/Log.kt
```diff
@@ -1,25 +1,141 @@
...
+import org.listenbrainz.android.BuildConfig
+import org.listenbrainz.shared.util.Log as sharedLog
+import java.io.File
+import java.io.FileWriter
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale.getDefault
+
+
+object Log {
```
07jasjeet: This class is not testable and its usages are not testable either. We need to inherit an interface here and provide that log interface via constructor parameters of a class that will use logging. You can keep it simple by just doing the following:

```kotlin
class XX(val logger: Logger = Log) 
```

app/src/main/java/org/listenbrainz/android/util/Utils.kt
```diff
@@ -680,48 +682,45 @@ object Utils {
         }
     }
 
-    fun submitLogs(context: Context) {
-        Logger.apply {
-            compressLogsInZipFile { zipFile ->
-                zipFile?.let {
-                    FileIntent
-                        .fromFile(
+    suspend fun submitLogs(context: Context) = withContext(Dispatchers.IO) {
```
07jasjeet: Should be part of Log implemention itself

app/src/main/java/org/listenbrainz/android/util/Log.kt
07jasjeet: Actually on another view, this class itself should not exist, we should use kermit logger that is present in shared module. Any platform specific implementation should exist in platform folders in shared module. We should directly reference the shared module logger in actual code.

Constructor approach still stands.

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/ConsentWebViewClient.kt
```diff
@@ -3,7 +3,7 @@ package org.listenbrainz.android.ui.screens.onboarding.auth
 import android.graphics.Bitmap
 import android.webkit.WebView
 import android.webkit.WebViewClient
-import com.limurse.logger.Logger
+import org.listenbrainz.android.util.Log
```
07jasjeet: kermit logger should be used here, provided via constructor.

07jasjeet (changes requested):
Need only one class/object for logging and it should be in shared. `submitLogs` should be inside shared module's logger impl as well such that it can be mocked where used and platform specific impl's can be used.

app/src/main/java/org/listenbrainz/android/util/Log.kt
nirvan73: I have passed the shared logger (kermit) via constructor in every classes its being used, thus making them testable :+1:

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/introduction/IntroductionScreens.kt
nirvan73: I have handled the scope via viewModelScope, but for the loader part as we click on the button to submit the logs the bottom sheet opens, thus blocks spam clicks but for the period while the logs are preparing, i have disabled the button for that period and enable once completed. let me know if u still prefer a separate loader with submitting info ??

nirvan73:
Hey @07jasjeet, I have made the requested changes, Please have a look them.

app/src/main/java/org/listenbrainz/android/di/KoinModules.kt
```diff
@@ -490,6 +494,28 @@ val appModule = module {
...
+        BuildInfo(
+            applicationId = BuildConfig.APPLICATION_ID,
+            versionCode = BuildConfig.VERSION_CODE,
+            versionName = BuildConfig.VERSION_NAME,
+            buildType = BuildConfig.BUILD_TYPE
+        )
+    }
+
+    single<co.touchlab.kermit.Logger> {
+        val context = androidContext()
```
07jasjeet: This should be provided via shared kmp module.
Currently this file exists in android dir but when its migrated to kmp, we would face errors here.

app/src/main/java/org/listenbrainz/android/di/KoinModules.kt
```diff
@@ -490,6 +494,28 @@ val appModule = module {
...
+        val logDirectory = context.getExternalFilesDir(null)?.path.orEmpty()
+        provideLogger(logDirectory, buildInfo = get())
+            .also {
+                Log.sharedLogger = it
+            }
+    }
+
+    single<LogSubmitter> {
+        AndroidLogSubmitter(get())
```
07jasjeet: Same, should be provided via kmp module.

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/createaccount/CreateAccountScreenLayout.kt
```diff
@@ -89,13 +90,15 @@ fun CreateAccountScreenLayout(
     onVerificationCompleteClick: () -> Unit = {},
     onPressBackInVerificationState: () -> Unit = {},
     onRefreshClick: () -> Unit = {},
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    submitLogs:(PlatformContext)->Unit
```
07jasjeet: Shouldn't need context here. Our LogSubmitter, which is platform specific, should use app context to write to files. Our submit logs function should be no-parameter call

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/auth/login/LoginScreenLayout.kt
```diff
@@ -154,7 +158,11 @@ fun LoginScreenLayout(
             modifier = Modifier
                 .statusBarsPadding()
                 .align(Alignment.TopEnd)
-                .padding(top = 8.dp, end = 8.dp)
+                .padding(top = 8.dp, end = 8.dp),
+            isSubmitting = uiState.isLogSubmitting,
+            submitLogs = {
+                submitLogs(context)
```
07jasjeet: Same here

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/introduction/IntroductionScreens.kt
```diff
@@ -331,17 +335,20 @@ fun OnboardingBackButton(modifier: Modifier = Modifier, onBackPress: (() -> Unit
 }
 
 @Composable
-fun OnboardingSupportButton(modifier: Modifier){
+fun OnboardingSupportButton(modifier: Modifier,isSubmitting:Boolean,submitLogs:(PlatformContext)->Unit){
```
07jasjeet: Same here

app/src/main/java/org/listenbrainz/android/viewmodel/CreateAccountViewModel.kt
```diff
@@ -55,6 +62,26 @@ class CreateAccountViewModel : ViewModel() {
 
     private var createAccountTimeoutJob: Job? = null
 
+    fun logSubmit(context: PlatformContext){
```
07jasjeet: Very very wrong, never pass context of activity to view model because a `viewModelScope` always has a longer lifecycle than context object. This code can cause memory leaks (someone holds context object's reference longer than its intended to)

app/src/main/java/org/listenbrainz/android/viewmodel/DashBoardViewModel.kt
```diff
@@ -77,6 +85,22 @@ class DashBoardViewModel(
         }
     }
 
+    fun logSubmit(context: PlatformContext){
```
07jasjeet: Same here.

shared/src/commonMain/kotlin/org/listenbrainz/shared/viewmodel/SettingsViewModel.kt
```diff
@@ -1,20 +1,47 @@
...
+    private val logger: Log = Log
 ): ViewModel() {
 
     private val _logoutStatus = MutableStateFlow<Boolean?>(null)
     val logoutStatus: StateFlow<Boolean?> = _logoutStatus.asStateFlow()
 
+    private val _submittingLogs = MutableStateFlow(false)
+    val submittingLogs: StateFlow<Boolean> = _submittingLogs.asStateFlow()
+
+    fun logSubmit(context : PlatformContext){
```
07jasjeet: Same here

shared/src/commonMain/kotlin/org/listenbrainz/shared/viewmodel/SettingsViewModel.kt
```diff
@@ -1,20 +1,47 @@
...
     val appPreferences: AppPreferences,
+    private val logSubmitter: LogSubmitter,
+    private val logger: Log = Log
 ): ViewModel() {
 
     private val _logoutStatus = MutableStateFlow<Boolean?>(null)
     val logoutStatus: StateFlow<Boolean?> = _logoutStatus.asStateFlow()
 
+    private val _submittingLogs = MutableStateFlow(false)
+    val submittingLogs: StateFlow<Boolean> = _submittingLogs.asStateFlow()
```
07jasjeet: Rather than having two flow value holders (one private and one public), I would rather advise using uiState approach we use over our app. (FeedViewModel or ProfileViewModel)

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/AndroidLogSubmitter.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.io.FileOutputStream
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
+
+actual interface LogSubmitter {
+    actual suspend fun submitLogs(context: PlatformContext)
+}
+
+class AndroidLogSubmitter(
+    private val buildConfig: BuildInfo
```
07jasjeet, line 18: should have app context provided from constructor

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/AndroidFileLogWriter.kt
```diff
@@ -0,0 +1,118 @@
...
+
+
+class AndroidFileLogWriter(
+    logDirectory: String,
+    private val buildConfig: BuildInfo,
+) : LogWriter(), SharedFileLogWriter {
+
+    private val dateFormat = SimpleDateFormat("dd-MM-yyyy-HH:mm:ss",Locale.getDefault())
+
+    private val loggerScope = CoroutineScope(Dispatchers.IO)
```
07jasjeet: Just use `GlobalScope` since log writing should only end when application dies.

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/AndroidFileLogWriter.kt
```diff
@@ -0,0 +1,118 @@
...
+                    sharedLog.log(Severity.Error, tag = "LogFileWriter", message =  "Failed to write to file", throwable =  e)
+                }
+            }
+            collectStartupData()
+        }catch (e:Exception){
+            sharedLog.log(Severity.Error, tag = "LogFileWriter", message =  "Failed to initialize file logging", throwable = e)
+        }
+    }
+
+    private fun collectStartupData() {
```
07jasjeet: We have build info from our DI coming in, use that and get rid of this here.

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/AndroidFileLogWriter.kt
```diff
@@ -0,0 +1,118 @@
...
+import kotlinx.coroutines.launch
+import java.io.File
+import java.io.FileWriter
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+import java.util.Locale.getDefault
+
+
+class AndroidFileLogWriter(
```
07jasjeet, line 11: I feel like more of this code could be put in commonMain because it has logic that can be re-used for all platforms. Channel usages, date format (Use kotlinx.dateTime), coroutine scopes, logging, try catching, severity to string mapping etc. We should just have things like FileWriter and file location here only.

shared/src/commonMain/kotlin/org/listenbrainz/shared/Platform.kt
```diff
@@ -1,3 +1,11 @@
...
\ No newline at end of file
+import co.touchlab.kermit.Logger
+import org.listenbrainz.shared.util.BuildInfo
+
+expect fun platform(): String
+
+expect fun provideLogger(
+    logFileDirectory:String?=null,
+    buildInfo: BuildInfo?
+): Logger
```
07jasjeet: `expect fun provideLogSubmitter` here

shared/src/iosMain/kotlin/org/listenbrainz/shared/util/IosLogSubmitter.kt
```diff
@@ -0,0 +1,16 @@
...
+
+actual interface LogSubmitter{
+    actual suspend fun submitLogs(context: PlatformContext)
+}
+
+class IosLogSubmitter(
+    private val buildConfig: BuildInfo
+): LogSubmitter{
+    override suspend fun submitLogs(context: PlatformContext) {
+        TODO("Not yet implemented")
```
07jasjeet, line 8: need implementation here.

07jasjeet (changes requested):

shared/src/iosMain/kotlin/org/listenbrainz/shared/util/IosLogSubmitter.kt
07jasjeet: Can do later, ignore

app/src/main/java/org/listenbrainz/android/di/KoinModules.kt
07jasjeet: Here, by `this`, I mean log directory.

shared/src/commonMain/kotlin/org/listenbrainz/shared/viewmodel/SettingsViewModel.kt
nirvan73: in just `SettingsViewModel` and `DashboardViewModel` i have used this approach `isSubmittingLogs` as there was no already existing common uiState for them, other than these in every other viewmodel like `LoginViewModel` i have passed it via uiState only.

I can do the uistate approach in `SettingsViewModel` but its quite tricky to do so in `DashboadViewModel` as its being served in multiple screens with no common uiState ??

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/AndroidFileLogWriter.kt
nirvan73: from the constructor, `buildInfo` is already being passed in as `buildConfig: BuildInfo` which is injected via DI and then the info is used here, should i remove the App version, is that what u meant ??

nirvan73:
Hey @07jasjeet I have made all the requested changes, please have a look and let me know if further changes are required.

app/src/main/java/org/listenbrainz/android/ui/screens/main/MainActivity.kt
```diff
@@ -273,7 +273,7 @@ class MainActivity : ComponentActivity() {
                 onboardingScreensQueue.add(NavigationItem.OnboardingScreens.LoginScreen)
             }
 
-            if (dashBoardViewModel.permissionStatusFlow.first()
+            if (dashBoardViewModel.uiState.value.permissionStatus
```
07jasjeet: these are two different things, one waits for a correct emission from preferences and assumes that the backed flow is not state flow, other is just straight up using whatever comes first. Maybe we should keep this original flow as is or add a special state on which we can call `.first { it != UNSET }` . I would still rather keep it the way it was.

shared/src/iosMain/kotlin/org/listenbrainz/shared/di/SharedAppModule.ios.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.listenbrainz.shared.repository.PlatformContext
+import org.listenbrainz.shared.util.BuildInfo
+import org.listenbrainz.shared.util.IosLogSubmitter
+import org.listenbrainz.shared.util.Log
+import org.listenbrainz.shared.util.LogSubmitter
+import platform.Foundation.NSDocumentDirectory
+import platform.Foundation.NSSearchPathForDirectoriesInDomains
+import platform.Foundation.NSUserDomainMask
+
+actual val platformModule = module {
```
07jasjeet: There need not be platform specific modules because both of these deps are provided in `Platform.kt`. Having one in shared should be enough. Move logDirectory to platform actual implementation of provideLogger. If we ever need to switch directory, we can also make provideLoggerDir same as how its done for data store kmp.

shared/src/commonMain/kotlin/org/listenbrainz/shared/di/SharedAppModule.kt
```diff
@@ -0,0 +1,5 @@
+package org.listenbrainz.shared.di
+
+import org.koin.core.module.Module
+
+expect val platformModule: Module
```
07jasjeet: should be actual rather than expect.

07jasjeet (changes requested):

07jasjeet (approved):

07jasjeet:
Thanks for the contribution @nirvan73 @javaman97 :)

PR #743, opened 2026-05-29
Added KMP lifecycle dependencies to shared module

nirvan73:
This PR adds the required libs in the shared module to support lifecycle in KMP.

### DIFF
--- a/shared/build.gradle.kts
+++ b/shared/build.gradle.kts
@@ -81,6 +81,8 @@ kotlin {
                 implementation(libs.androidx.lifecycle.runtime)
                 // koin
                 implementation(libs.koin.core)
+                implementation(libs.koin.compose)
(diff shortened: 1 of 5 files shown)


shared/build.gradle.kts
```diff
@@ -76,6 +76,13 @@ kotlin {
                 implementation(libs.datastore.preferences.core)
                 implementation(libs.kotlinx.serialization.json)
                 api(libs.kotlinx.coroutines.core)
+                // lifecycle
+                api(libs.androidx.lifecycle.viewmodel)
```
07jasjeet, line 80: shouldn't these android deps be in androidMain and not commonMain?

nirvan73: image

after this version it supports KMP (we are using 2.10.0) and it configures the lifecycle for each platform itself

PR #752, opened 2026-06-14
Migrated BrainzPlayer Database to shared module

nirvan73:
This PR migrates `BrainzPlayerDatabase` from `androidApp` module to shared module

## Key Changes:- 
1. Moved `TypeConvertor` to shared module with all the related DAOs and their entities.
2. Moved `BrainzPlayerDatabase` to shared module while registering it to `sharedDatabaseModule` with platform specific providers
3. Registered all the DAOs in `sharedDaoModule`.
4. Moved `Transformer` to shared module and all the related data classes with it.
5. Added Room keep rule in `progaurd-rules.pro` file to prevent shrinking crashes.
6. Wired all the shared modules inside of app module level `KoinModules`.


shared/src/commonMain/kotlin/org/listenbrainz/shared/util/TypeConverter.kt
```diff
@@ -0,0 +1,114 @@
...
+        dayOfMonth(Padding.ZERO)
+        char(',')
+        char(' ')
+        amPmHour(Padding.NONE)
+        char(':')
+        minute(Padding.ZERO)
+        char(' ')
+        amPmMarker("AM","PM")
+    }): String {
+        val instant = Instant.fromEpochMilliseconds(microSeconds/1000)
```
07jasjeet: This flips the old `* 1000` into `/ 1000`, but the caller in FeedEventType passes seconds, so feed timestamps older than a day will render as ~1970. Should stay `* 1000`.

shared/src/commonMain/kotlin/org/listenbrainz/shared/model/Playlist.kt
```diff
@@ -11,21 +9,21 @@ data class Playlist(
     val id: Long = Random.nextLong(),
     val title: String = "",
     val items: List<Song> = listOf(),
-    @DrawableRes val art: Int = R.drawable.ic_queue_music
+    val art: String = "ic_queue_music"
```
07jasjeet, line 12: Moving `art` from a drawable res Int to a String breaks PlaylistScreen (`model = it.art`), Coil can't resolve `ic_liked`/`ic_queue_music` so the built-in playlists fall back to the error icon. We'll need a name->res mapping in the UI.

shared/src/commonMain/kotlin/org/listenbrainz/shared/di/SharedDaoModule.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.listenbrainz.shared.model.dao.AlbumDao
+import org.listenbrainz.shared.model.dao.ArtistDao
+import org.listenbrainz.shared.model.dao.PlaylistDao
+import org.listenbrainz.shared.model.dao.SongDao
+
+val sharedDatabaseModule = module {
+    single<BrainzPlayerDatabase> {
+        getBrainzPlayerDatabase(get())
+            .setDriver(BundledSQLiteDriver())
+            .fallbackToDestructiveMigration(dropAllTables = true)
```
07jasjeet: Bumping to v3 with destructive fallback and no 2->3 migration wipes existing users' playlists and play history on upgrade, and it leaves MIGRATION_1_2 as dead code. Is that intended?

shared/build.gradle.kts
```diff
@@ -28,6 +28,14 @@ kotlin {
         }.configure {
             instrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
         }
+
+        optimization {
+            minify = true
+            consumerKeepRules.apply {
+                publish = true
+                file("progaurd-rules.pro")
```
07jasjeet: Filename's misspelled and this file doesn't exist under shared/, and `file(...)` inside apply is a no-op anyway, so with minify on the Room consumer keep rules won't actually get applied.

shared/src/iosMain/kotlin/org/listenbrainz/shared/Platform.ios.kt
```diff
@@ -13,6 +13,14 @@ import platform.Foundation.NSDocumentDirectory
 import platform.Foundation.NSSearchPathForDirectoriesInDomains
 import platform.Foundation.NSUserDomainMask
 
+import androidx.room.Room
+import androidx.room.RoomDatabase
+import kotlinx.cinterop.ExperimentalForeignApi
+import org.listenbrainz.shared.di.database.BrainzPlayerDatabase
+import platform.Foundation.NSDocumentDirectory
```
07jasjeet: NSDocumentDirectory and NSUserDomainMask are already imported a few lines up, these are duplicates.

shared/src/commonMain/kotlin/org/listenbrainz/shared/util/TypeConverter.kt
```diff
@@ -0,0 +1,114 @@
+package org.listenbrainz.shared.util
+
+import androidx.compose.ui.text.intl.Locale
+import kotlinx.datetime.Instant
```
07jasjeet: kotlinx.datetime.Instant is deprecated in 0.7.x, SongRepository already uses kotlin.time.Clock so might as well keep this consistent.

07jasjeet (commented):
Nice migration overall, the expect/actual DB setup reads clean. A few things worth a look before merge, mostly the two runtime regressions inline below. Also worth confirming DaoTest still runs under the new KMP Room setup since it doesn't call setDriver(BundledSQLiteDriver()).

shared/src/commonMain/kotlin/org/listenbrainz/shared/di/SharedDaoModule.kt
07jasjeet: The app should always crash in case of faulty migration.

nirvan73: For the destructive fallback and the version 3, i added these while testing it locally and since i was making changes in the `PlaylistEntity` schema, so will remove the fallback, but for the version 2->3 migration for now i will register a `MIGRATION_2_3` script to handle this schema update for the datatype of the art variable under Playlists, should i go with this ??

07jasjeet: Yes

07jasjeet: Please take migrations seriously next time on. We cannot afford local testing changes to be pushed for review. Also, there shouldn't be any need for it in the first place. Always try to make do with migrations.

PR #766, opened 2026-07-04
Removed Brainz-Player 

nirvan73:
This PR removes BrainzPlayer feature completely from the ListenBrainz-android app.


app/src/main/java/org/listenbrainz/android/ui/screens/main/HomeScreen.kt
```diff
@@ -143,36 +113,23 @@ fun HomeScreen(
...
-                }
+            navController.navigate(AppNavigationItem.SearchScreen.route) {
+                launchSingleTop = true
+                restoreState = true
             }
         }
     )
     val navOrder = dashboardUiState.navBarOrder
 
     val filteredNavItems = navOrder?.filter {
```
07jasjeet: This filter no longer looks at the item, so it's all-or-nothing on the permission: on API 33+ with READ_MEDIA_AUDIO denied, `filteredNavItems` comes back empty, `startRoute` is null, and the `startRoute != null` check below means the nav graph and backdrop never render at all. Blank app. Since the player is gone this can just be `navOrder`.

app/src/main/java/org/listenbrainz/android/ui/screens/main/HomeScreen.kt
```diff
@@ -110,26 +99,7 @@ fun HomeScreen(
...
-
-    val isNothingPlaying = remember(currentlyPlayingSong) {
-        currentlyPlayingSong.toSong.title == "null"
-                && currentlyPlayingSong.toSong.artist == "null"
-                || currentPlayableState.songs.isEmpty()
-    }
-
-    val isListeningNowOpenedInConcealedState = backdropScaffoldState.targetValue != BackdropValue.Revealed && isNothingPlaying && listeningNowUIState.isListeningNow
+    val isListeningNowOpenedInConcealedState = backdropScaffoldState.targetValue != BackdropValue.Revealed && listeningNowUIState.isListeningNow
     val isAudioPermissionGranted = permissions[PermissionEnum.ACCESS_MUSIC_AUDIO] == PermissionStatus.GRANTED || !PermissionEnum.ACCESS_MUSIC_AUDIO.isPermissionApplicable()
```
07jasjeet: And once the filter above is gone, nothing reads local audio anymore, so this whole gate can go with it. Same for `ACCESS_MUSIC_AUDIO` / `READ_EXTERNAL_STORAGE` in PermissionEnum and `READ_MEDIA_AUDIO` + `FOREGROUND_SERVICE_MEDIA_PLAYBACK` in the manifest. Better to drop them here than keep asking users for access we don't use.

app/src/main/java/org/listenbrainz/android/ui/screens/search/BrainzPlayerSearchScreen.kt
```diff
@@ -1,91 +0,0 @@
-package org.listenbrainz.android.ui.screens.search
```
07jasjeet, line 1: This one got emptied instead of deleted, the 0-byte file is still tracked. `git rm` it?

app/src/main/java/org/listenbrainz/android/ui/screens/listeningNow/ListeningNowBackDropScreen.kt
```diff
@@ -0,0 +1,159 @@
...
+        frontLayerBackgroundColor = defaultBackgroundColor,
+        appBar = {},
+        persistentAppBar = false,
+        frontLayerContent = {
+            LaunchedEffect(Unit) {
+                val delta = backdropScaffoldState.offsetOrZero()
+                maxDelta = max(delta, maxDelta)
+            }
+
+            LaunchedEffect(listeningNowUIState.song) {
```
07jasjeet, line 98: The old version had a reveal guard in the other branch here (`if (isNothingPlaying && isConcealed) reveal()`, with the comment "to prevent screen showing null abruptly after listening now finishes"). Now the front layer always renders ListeningNowScreen, and ListeningNowLayout draws only the gradient when `song == null`. So if listening-now expires while the screen is open fullscreen, the user is left on an empty screen. Worth revealing when the song goes null?

app/src/main/java/org/listenbrainz/android/ui/screens/listeningNow/ListeningNowBackDropScreen.kt
```diff
@@ -0,0 +1,159 @@
...
+    } catch (_: IllegalStateException) {
+        0f
+    }
+
+@ExperimentalMaterialApi
+@Composable
+fun ListeningNowBackDropScreen(
+    modifier: Modifier = Modifier,
+    backdropScaffoldState: BackdropScaffoldState,
+    listeningNowViewModel: ListeningNowViewModel = viewModel(),
```
07jasjeet: `viewModel()` can't construct a Koin VM with constructor args, and everything else in the app uses `koinViewModel()`. HomeScreen always passes one in so it never bites in practice, but the preview at the bottom calls this without one. Carried over from the old file, but worth switching while it's a fresh file.

app/src/main/java/org/listenbrainz/android/ui/screens/listeningNow/ListeningNowBackDropScreen.kt
```diff
@@ -0,0 +1,159 @@
...
+    isLandscape: Boolean = LocalConfiguration.current.orientation == Configuration.ORIENTATION_LANDSCAPE,
+    backLayerContent: @Composable () -> Unit
+) {
+    var maxDelta by rememberSaveable {
+        mutableFloatStateOf(0F)
+    }
+    val defaultBackgroundColor = ListenBrainzTheme.colorScheme.background
+    val listeningNowUIState by listeningNowViewModel.listeningNowUIState.collectAsState()
+
+    val isListeningNow = remember(listeningNowUIState) {
```
07jasjeet: This is the only remaining user of the local `isListeningNow`, while line 95 uses `listeningNowUIState.isListeningNow` (which is just `song != null`). Two slightly different definitions of the same thing in one file, could we settle on the uiState one?

app/src/main/java/org/listenbrainz/android/ui/screens/listeningNow/ListeningNowBackDropScreen.kt
```diff
@@ -0,0 +1,159 @@
...
+                )
+            }
+
+        })
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Preview
+@Composable
+fun BrainzPlayerBackDropScreenPreview() {
```
07jasjeet: Stale name: `ListeningNowBackDropScreenPreview`.

app/src/main/java/org/listenbrainz/android/di/KoinModules.kt
```diff
@@ -242,32 +230,12 @@ val repositoryModule = module {
     // API Repositories
     single<YimRepository> { YimRepositoryImpl(get()) }
     single<Yim23Repository> { Yim23RepositoryImpl(get()) }
-    single<AppUpdatesRepository> { AppUpdatesRepositoryImpl(get(), get(), get(named(IO_DISPATCHER))) }
-
-}
-
-// Service Module for BrainzPlayer
-val playerModule = module {
```
07jasjeet, line 250: playerModule is gone but the ExoPlayer deps are still declared: `google.exoplayer.core/ui/mediasession` in app/build.gradle.kts and the `exoplayer = "2.19.1"` version in libs.versions.toml. Nothing uses them now, good chance to drop them.

shared/src/commonMain/kotlin/org/listenbrainz/shared/util/TypeConverter.kt
```diff
@@ -25,30 +23,6 @@ object TypeConverter {
         return json.decodeFromString(playableJSON)
```
07jasjeet: These two lost their last consumer as well. Same for `Playable`/`PlayableType`, `Song`, `Playlist`, `AppPreferences.currentPlayable`, `SearchData.Songs`, `CustomSeekBar` in SeekBar.kt, and `SongsData`/`AlbumsData` (still registered in SharedAppModule, and the Android impl is what pulls in the MediaStore audio query). Also the leftover strings (`brainzplayer_notification_channel_*`) and the BrainzPlayer onboarding slide in FeaturesActivity / IntroScreenDataEnum. One more sweep and "completely" really is complete.

07jasjeet (commented):
Mostly looks good, a few things to address

app/src/main/java/org/listenbrainz/android/ui/screens/main/HomeScreen.kt
```diff
@@ -110,27 +99,7 @@ fun HomeScreen(
         }
     }
```
07jasjeet, line 100: This maxOffset/playerHeight/isBackdropInitialised block only fed the removed player background color, so it can go too. Same for the unused `permissions` on line 66.

shared/src/commonMain/kotlin/org/listenbrainz/shared/repository/AppPreferences.kt
```diff
@@ -50,14 +47,6 @@ interface AppPreferences {
     val shouldListenNewPlayers: DataStorePreference<Boolean>
 
     val isNotificationServiceAllowed: Boolean
```
07jasjeet, line 49: ALBUMS_ON_DEVICE, SONGS_ON_DEVICE and CURRENT_PLAYABLE in PreferenceKeys.kt have no readers left, remove those too?

shared/src/androidMain/kotlin/org/listenbrainz/shared/Platform.android.kt
```diff
@@ -67,15 +61,6 @@ actual fun provideRemotePlaybackHandler(
...
-
-actual fun getBrainzPlayerDatabase(): RoomDatabase.Builder<BrainzPlayerDatabase> {
-    val brainzPlayerDb = applicationContext.getDatabasePath("brainzplayer_database")
-    return Room.databaseBuilder<BrainzPlayerDatabase>(
-        context = applicationContext,
-        name = brainzPlayerDb.absolutePath
-    )
-}
-
 actual fun provideListensRepositoryImpl(
```
07jasjeet, line 64: Upgrading users will keep the old brainzplayer_database file on disk forever. Worth a one-time context.deleteDatabase("brainzplayer_database") on startup?

app/src/main/AndroidManifest.xml
```diff
@@ -63,15 +61,6 @@
                 <action android:name="android.service.notification.NotificationListenerService" />
             </intent-filter>
         </service>
```
07jasjeet, line 63: Existing installs will still show the "BrainzPlayer Now Playing" channel in system notification settings, should we deleteNotificationChannel on upgrade?

app/src/main/res/values/strings.xml
```diff
@@ -207,8 +207,6 @@
     <string name="title_current_track">Current Track</string>
     <string name="title_image_size">Image size</string>
     <string name="title_image_scale_type">Scale type</string>
```
07jasjeet, line 209: A few resources are orphaned now: music_player.json, intro_speakers.png, ic_audio_file, ic_storage, player_selected/unselected (still mapped in AppNavigationItemExt) and bp_loading_text in all locales. Fine to sweep here or in a follow-up.

app/src/main/java/org/listenbrainz/android/util/HeaderInterceptor.kt
```diff
@@ -18,13 +17,13 @@ class HeaderInterceptor (
         var request: Request = chain.request()
         
         return runBlocking {
-            if (request.headers[AUTHORIZATION].isNullOrEmpty()) {
+            if (request.headers("Authorization").isNullOrEmpty()) {
```
07jasjeet, line 20: request.header("Authorization") (singular) keeps the old nullable String semantics, the list overload is never null so isNullOrEmpty reads odd here.

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/FeaturesActivity.kt
```diff
@@ -37,27 +37,13 @@ class FeaturesActivity : OnboardAdvanced() {
...
-                )
-            },
-            slideNumber = 1,
-            required = true
-        )
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            askForPermissions(
+                arrayOf(Manifest.permission.POST_NOTIFICATIONS),
+                slideNumber = 1,
+                required = false
```
07jasjeet, line 44: required went from true to false here, intentional?

07jasjeet (commented):
Nice cleanup, a few small things to address

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/FeaturesActivity.kt
nirvan73: Is there a specific reason POST_NOTIFICATIONS was marked as required ?

07jasjeet (commented):
LGTM, nice work - everything from the earlier rounds is addressed, just a few tiny leftovers

app/src/main/java/org/listenbrainz/android/ui/screens/main/HomeScreen.kt
```diff
@@ -49,16 +46,11 @@ import org.listenbrainz.android.ui.navigation.AdaptiveNavigationBar
 import org.listenbrainz.android.ui.navigation.AppNavigation
 import org.listenbrainz.android.ui.navigation.NavBarReorderOverlay
 import org.listenbrainz.android.ui.navigation.TopBarActions
-import org.listenbrainz.android.ui.screens.brainzplayer.BrainzPlayerBackDropScreen
+import org.listenbrainz.android.ui.screens.listeningNow.ListeningNowBackDropScreen
 import org.listenbrainz.android.ui.screens.onboarding.permissions.PermissionEnum
```
07jasjeet: A few imports lost their users with the player state gone: collectAsState (26), derivedStateOf (27), mutableFloatStateOf (29), PermissionStatus (44), PermissionEnum (50), toPx (53). The commented-out line 161 can go too.

app/src/main/java/org/listenbrainz/android/ui/screens/listeningNow/ListeningNowBackDropScreen.kt
```diff
@@ -0,0 +1,163 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.RectangleShape
+import androidx.compose.ui.graphics.graphicsLayer
+import androidx.compose.ui.platform.LocalConfiguration
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.util.lerp
+import androidx.lifecycle.viewmodel.compose.viewModel
```
07jasjeet: remember (21) and this viewModel import are unused now that the default is koinViewModel().

app/src/main/java/org/listenbrainz/android/ui/screens/listeningNow/ListeningNowBackDropScreen.kt
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    /** 56.dp is default bottom navigation height */
+    val headerHeight by animateDpAsState(
+        targetValue = if (isLandscape) 0.dp else
+            if (!listeningNowUIState.isListeningNow)
+                56.dp
+            else
+                56.dp + ListenBrainzTheme.sizes.brainzPlayerPeekHeight
```
07jasjeet: Last naming leftovers: brainzPlayerPeekHeight here plus the brainzPlayer gradient brushes and color comments in Theme.kt / colors.xml. Fine to rename in a follow-up since they are still in use.

app/src/main/java/org/listenbrainz/android/application/App.kt
```diff
@@ -42,10 +43,23 @@ class App : Application(), Configuration.Provider {
         }
 
         GlobalScope.launch {
+            deleteBrainzPlayerDatabaseIfExists()
             startListenService(appPreferences)
         }
     }
 
+    private fun deleteBrainzPlayerDatabaseIfExists(){
```
07jasjeet: Nit: this also clears the old notification channel, so the name undersells it a bit. cleanupBrainzPlayerRemnants or similar? The "Music" id matches the old NOTIFICATION_CHANNEL_ID, nicely done.

app/src/main/java/org/listenbrainz/android/ui/screens/onboarding/FeaturesActivity.kt
07jasjeet, line 44: To answer the question from the earlier thread: required=true was there for the media/storage permissions that used to be bundled in this call, which the player genuinely needed. POST_NOTIFICATIONS alone should not block onboarding, so false is the right call.

07jasjeet (approved):

PR #767, opened 2026-07-18
Implemented Shared Notification Service, Shared WorkManager and migrated Listens submission to shared module

nirvan73:
#### Rebased with latest upstream/cmp, this branch contains every commits of the base PR.

---

This PR implements a `Platform Notification Manager` (shared notification manager) and shared work manager [KmpWorkManager](https://github.com/brewkits/kmpworkmanager). This PR also migrates the Listen submission feature to shared module under `androidMain`.
### Key Changes:- 
1. Created `AppArrayProvider` for accessing the array resources in shared module for now.
2. Modified `AppDrawableProvider` and `AppStringProvider` acc to the need.
3. Added the kmp work manager dependencies under shared module.
4. Implemented `PlatformNotificationManager` with platform specific impl.
5. Replaced android work manager with kmp work manager and updated `ListenSubmissionWorker` acc.
6. Moved to `androidMain` and updated `ListenSubmissionState` to use shared notification manager and shared work manager.
7. Moved all necessary data models and utility functions to `androidMain`.
8. Registered the shared notification manager under `sharedAppModule` via platform specific providers.
9. Initialized KMP WorkManager in `App` and removed native configurations
10. Updated `KoinModules` to register the modified `ListenServiceManager`.


07jasjeet (commented):
Nice migration, mostly looks good. A few things to check.

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/ListenSubmissionState.kt
```diff
@@ -0,0 +1,248 @@
...
+
+    private fun submitPlayingNow() {
+        if (!playingTrack.playingNowSubmitted) {
+            scope.launch {
+                scheduler.enqueue(
+                    id = "playing-now-${playingTrack.pkgName}",
+                    trigger = TaskTrigger.OneTime(initialDelayMs = 0),
+                    workerClassName = "ListenSubmissionWorker",
+                    inputJson = Json.encodeToString(
+                        ListenWorkerInput(playingTrack, ListenType.PLAYING_NOW)
```
07jasjeet: scope.launch runs on IO, but playingTrack is a var that onNewMetadata reassigns. By the time this coroutine actually runs, playingTrack may already point at a newer track, so we could serialize the wrong one. Safer to grab `val track = playingTrack` before the launch and use that (same in submitListen).

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/ListenSubmissionState.kt
```diff
@@ -0,0 +1,248 @@
...
+    // Utility functions
+    
+    private fun roundDuration(duration: Long): Long =
+        (duration / 1000) * 1000
+
+    private fun submitPlayingNow() {
+        if (!playingTrack.playingNowSubmitted) {
+            scope.launch {
+                scheduler.enqueue(
+                    id = "playing-now-${playingTrack.pkgName}",
```
07jasjeet: What does scheduler.enqueue do when the same id is enqueued twice, replace/keep/append? playing-now is keyed only by pkgName, so rapid updates from the same player all share one id. The old WorkManager.enqueue always created fresh work, so just want to be sure we're not silently dropping submissions.

gradle/libs.versions.toml
```diff
@@ -63,7 +63,7 @@ composeRatingbar = "1.3.12"
 loggerAndroid = "1.0.0"
 compileSdk = "36"
 targetSdk = "36"
-minSdk = "23"
+minSdk = "26"
```
07jasjeet, line 66: minSdk 23 -> 26 drops Android 6 / 7 / 7.1. Intentional (kmp-workmanager requirement?) or a side effect? Worth calling out since it drops real users.

shared/src/androidMain/kotlin/org/listenbrainz/shared/Platform.android.kt
```diff
@@ -77,4 +83,19 @@ actual fun getListensSubmissionDatabase(): RoomDatabase.Builder<ListensSubmissio
...
     )
+}
+
+actual fun provideSharedNotificationManager(
+    drawableProvider: DrawableProvider,
+    stringProvider: StringProvider,
+    arrayProvider: ArrayProvider
+): PlatformNotificationManager {
+
+    val targetClass = NotificationConfig.targetActivityClass ?: Class.forName("android.app.Activity")
```
07jasjeet: The android.app.Activity fallback would create an untappable notification if targetActivityClass is ever null. Application.onCreate should set it first in practice, but since there's no sensible default here, failing fast might be safer than silently pointing at a non-launchable class.

shared/src/commonMain/kotlin/org/listenbrainz/shared/util/PlatformNotificationManager.kt
```diff
@@ -0,0 +1,9 @@
+package org.listenbrainz.shared.util
+
+import org.listenbrainz.shared.model.PlayingTrack
+
+interface PlatformNotificationManager {
+    fun createChannel()
+    fun postListeningNotification(track: PlayingTrack?): Any
```
07jasjeet: Returning Any just so the Android side can hand a Notification back to startForeground is a bit leaky (and the caller does `as Notification`). Could keep this returning Unit and expose the foreground notification via an Android-only method instead.

app/src/main/java/org/listenbrainz/android/service/ListenSubmissionService.kt
```diff
@@ -127,32 +125,19 @@ class ListenSubmissionService(
...
-            nm?.deleteNotificationChannel(CHANNEL_ID)
-        }
+        notificationManager.deleteChannel()
     }
 
     var isStarted = false
+    @RequiresPermission(Manifest.permission.POST_NOTIFICATIONS)
     fun startForeground() {
-        val notification = getListeningNotification(null)
+        val notification = notificationManager.postListeningNotification(null) as Notification
```
07jasjeet: postListeningNotification also calls manager.notify internally, so this both posts notification 420 and then hands it to startForeground with the same id. Old code only built it for the foreground service. Probably fine since the id matches, but worth confirming it isn't double-posting.

gradle/libs.versions.toml
nirvan73: yes it is kmp-workmanager requirement and is declared in the library only as minSdkVersion cannot be smaller than version 26.

shared/src/commonMain/kotlin/org/listenbrainz/shared/util/PlatformNotificationManager.kt
```diff
@@ -0,0 +1,9 @@
+package org.listenbrainz.shared.util
+
+import org.listenbrainz.shared.model.PlayingTrack
+
+interface PlatformNotificationManager {
+    fun createChannel()
+    fun postListeningNotification(track: PlayingTrack?)
```
07jasjeet, line 7: This a specialized use-case and should not be part of generic notication manager. It can be a static extension function `PlatformNotificationManager.postListeningNotification(track: PlayingTrack?)` but not part of base implementation.

shared/src/iosMain/kotlin/org/listenbrainz/shared/util/PlatformUtils.ios.kt
```diff
@@ -1,9 +1,32 @@
...
-    actual fun getSHA1(context: PlatformContext,packageName:String):String?{
+actual object PlatformUtils {
+    actual fun getSHA1(context: PlatformContext, packageName: String): String? {
         return null
     }
+    actual suspend fun canShowNotifications(): Boolean {
+        return suspendCancellableCoroutine{
+            val center = UNUserNotificationCenter.currentNotificationCenter()
+            center.getNotificationSettingsWithCompletionHandler { settings->
+                if(!it.isActive){
```
07jasjeet, line 19: No need for this check, on resume, if our calling coroutine is already cancelled, cancellation exception will be thrown at call site.

shared/src/iosMain/kotlin/org/listenbrainz/shared/util/PlatformUtils.ios.kt
```diff
@@ -1,9 +1,32 @@
...
+                    return@getNotificationSettingsWithCompletionHandler
+                }
+                if(settings!=null){
+                    val status = settings.authorizationStatus
+                    val isGranted = status == UNAuthorizationStatusAuthorized || status == UNAuthorizationStatusProvisional || status == UNAuthorizationStatusEphemeral
```
07jasjeet, line 24: This whole call can be reduced to just:
```kotlin
val status = settings?.authorizationStatus
cont.resume(
    status == UNAuthorizationStatusAuthorized ||
    status == UNAuthorizationStatusProvisional ||
    status == UNAuthorizationStatusEphemeral
)
```

shared/src/commonMain/kotlin/org/listenbrainz/shared/model/PlayingTrack.kt
```diff
@@ -81,19 +71,11 @@ data class PlayingTrack(
...
-                releaseName = extractReleaseName(),
-                pkgName = pkgName
-            )
-        }
     }
-}
\ No newline at end of file
+}
+
+expect fun PlayingTrack.toSimilarTo(other: Any): Boolean
```
07jasjeet, line 81: Since we can convert MediaMetadata to PlayingTrack, lets get rid of this expect declaration and funnel all similar calls with PlayingTrack. Currently I can see MediaMetadata is never even passed, so we can change the original function signature to `PlayingTrack.toSimilarTo(track: PlayingTrack): Boolean`

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/NotificationConfig.kt
```diff
@@ -0,0 +1,11 @@
+package org.listenbrainz.shared.util
+
+import kotlin.reflect.KClass
+
+object NotificationConfig {
```
07jasjeet, line 5: This is very wrong. I can see we are just passing the MainActivity class using this. This field is exposed to the whole world and anyone in future can put in a sneaky assignment which can be missed in reviews and we'd have a bug on our hand that we cannot debug.

Also, if any case we may have native UIs where we might create another activity on android side, this solution will fail because the actual activity class might differ.

This class should be passed by callers themselves. Now this may be platform specific, so we might need a generic interface which will contain as much as common implementation as possible that may be casted/parsed to platform implementation if we need specifics. I feel this can be perfected better if we have an iOS app but still a rough vision should look like what I said. We need to leave a TODO on it to improve the API. 

Now why am I blabbering all this: In the age of AI, memorizing platform specific things is getting redundant very quickly. Things like API design, architecture, implementation loop and code quality is something that is better to learn rather than memorizing readily available facts.

shared/src/androidMain/kotlin/org/listenbrainz/shared/util/ListenSubmissionState.kt
```diff
@@ -0,0 +1,250 @@
...
+import org.listenbrainz.shared.model.PlayingTrack
+import org.listenbrainz.shared.model.TimerState
+import org.listenbrainz.shared.repository.PlatformContext
+import org.listenbrainz.shared.util.ListenSubmissionStateConstants.DEFAULT_DURATION
+import org.listenbrainz.shared.util.ListenSubmissionStateConstants.MAX_SUBMISSION_DURATION
+import org.listenbrainz.shared.util.ListenSubmissionStateConstants.SUBMISSION_TIMER_TOKEN
+import org.listenbrainz.shared.util.ListenSubmissionStateConstants.TRACK_COMPLETION_TIMER_TOKEN
+import org.listenbrainz.shared.util.PlatformUtils.canShowNotifications
+
+open class ListenSubmissionState : KoinComponent {
```
07jasjeet, line 28: Commit history is lost for this file...

